PR #3, opened 2022-04-14
Added a custom metric for extracting same-origin links

pmeenan:
This is the first bit of the logic for crawling beyond the main landing pages.
* extracts any links on images or H[1-6] tags that are in the viewport
* First checks for a child element with a link (useful for H[1-6] tags where the link is inside)
* Falls back to a parent element with a link (useful for images that are in a link or links that enclose divs with images/headings)
* aggregates the element sizes for elements that link to the same url
* returns a list of same-origin links sorted by size (from largest to smallest)

This has worked REALLY well at extracting the key content to crawl for all of the pages I have tested.

Sample WPT test: https://www.webpagetest.org/result/220415_AiDcEN_95B/

Sample output:

```json
"_crawl_links": [
  "https://www.cnn.com/europe/live-news/ukraine-russia-putin-news-04-15-22/h_3b2b55af52fd29581ecf9670ea794f1b",
  "https://www.cnn.com/2022/04/15/politics/mike-lee-chip-roy-text-messages-jan-6-mark-meadows-overturn-election/index.html",
  "https://www.cnn.com/videos/world/2022/04/15/jerusalem-clashes-israel-palestinians-gold-intl-vpx.cnn",
  "https://www.cnn.com/2022/04/14/opinions/finland-nato-russia-invasion-heinonen/index.html",
  "https://www.cnn.com/2022/04/14/politics/dustin-thompson-january-6-trump/index.html",
  "https://www.cnn.com/2022/04/14/health/covid-breath-test/index.html",
  "https://www.cnn.com/audio/podcasts/axe-files/episodes/e9e3ed5d-9747-4e59-993b-ae7700256371",
  "https://www.cnn.com/europe/live-news/ukraine-russia-putin-news-04-15-22/index.html",
  "https://www.cnn.com/style/article/alia-bhatt-ranbir-kapoor-wedding-intl-scli/index.html",
  "https://www.cnn.com/2022/04/14/uk/harry-megan-visit-queen-gbr-uk-intl/index.html",
  "https://www.cnn.com/cnn-underscored/gifts/amazon-easter-gifts",
  "https://www.cnn.com/2022/04/14/tech/elon-musk-twitter-offer/index.html"
],
```

Progress on https://github.com/HTTPArchive/data-pipeline/issues/12


rviscomi:
@tunetheweb could you help disable the linter error `Parsing error: 'return' outside of function`? That is standard procedure for custom metrics.

dist/sameorigin_links.js
```diff
@@ -0,0 +1,71 @@
...
+    uri2 = new URL(uri2);
+    if(uri1.host !== uri2.host) return false;
+    if(uri1.port !== uri2.port) return false;
+    if(uri1.protocol !== uri2.protocol) return false;
+    return true;
+}
+let links = {};
+const elementTypes = ['img', 'h1', 'h2', 'h3', 'h4', 'h5', 'h6'];
+const documentOrigin = window.location.href;
+for (let eType of elementTypes) {
```
rviscomi, line 37: This outer loop could be replaced with a more complex querySelectorAll:

```js
let elements = document.querySelectorAll('img, h1, h2, h3, h4, h5, h6');
```

dist/sameorigin_links.js
```diff
@@ -0,0 +1,71 @@
...
+    let elements = document.querySelectorAll(eType);
+    for (let e of elements) {
+        const rect = getBoundingClientRect(e);
+        if (rect.width > 1 && rect.height > 1 && intersectRect(rect, viewport)) {
+            let link = e.querySelector('a[href]') || e.closest('a[href]');
+            if (link && sameOrigin(link.href, documentOrigin)) {
+                let entry = {
+                    href: link.href,
+                    rect: rect,
+                    size: rect.width * rect.height
```
rviscomi, line 47: Should we do something with this metadata, perhaps as a way to sort the results by importance? If we can only crawl the top 2 or 3 secondary pages, this would be a useful way to choose.

dist/sameorigin_links.js
```diff
@@ -0,0 +1,71 @@
...
+                };
+                const title = link.getAttribute('title');
+                if (title) {
+                    entry['title'] = title;
+                }
+                const innerText = e.innerText;
+                if (innerText) {
+                    entry['text'] = innerText;
+                }
```
rviscomi, line 56: What did you have in mind for this metadata?

rviscomi (commented):
Nice! I had started a [branch](https://github.com/HTTPArchive/custom-metrics/blob/crawlable-urls/dist/crawlable_urls.js) with an LCP-only implementation, but I like this general img/heading approach better.

dist/sameorigin_links.js
pmeenan: It could, but I'm storing them under different indexes and sorting each by size so it was cleaner to keep it as a separate loop, otherwise the sorting will need its own loop and we'll need to extract the element type from each result.

pmeenan:
How about simplifying it to just produce an ordered list of URLs to pick from when crawling?
* No metadata reported about each link
* When a link is the same for different elements, join them, summing the size
* Query the img, h* elements without separating the results
* Reverse-sort by size, links that cover the largest aggregate area at the top
* Rename the metric to crawl_links

pmeenan:
Simplified the logic and updated the example in the commit description

pmeenan:
FYI, I disabled JAVASCRIPT_STANDARD but kept the eslint configs to get rid of the return error. If it makes sense to keep both linters then we can see if there is a way to disable just that one error within JAVASCRIPT_STANDARD

tunetheweb:
> Disable the javascript_standard linter (keep eslint)

We only use ESLint in other projects. Standard JS is not "standard" - the name always annoyed me. 😠

You may be better explicitly only _enabling_ the linters you want (that's what we do in the other HTTP Archive repos), rather than enabling all and explicitly _disabling_ the ones you don't want. That would be done with this (presuming it's only going to be JS and Markdown in this repo):

```
        env:
          VALIDATE_JAVASCRIPT_ES: true
          VALIDATE_MARKDOWN: true
```

Also you may want to copy this config file for the Markdown linting to this repo: https://github.com/HTTPArchive/legacy.httparchive.org/blob/master/.github/linters/.markdown-lint.yml.
For ES Lint we just use the defaults, so no config file.

pmeenan:
@tunetheweb thanks, switched to opt-in and picked up the markdown config

.github/linters/.markdown-lint.yml
```diff
@@ -0,0 +1,38 @@
...
+MD029: false                  # Ordered list item prefix
+MD033: false                  # Allow inline HTML
+MD034: false                  # Allow base URLs
+MD036: false                  # Emphasis used instead of a heading
+MD040: false                  # Don't demand language for all code blocks
+
+#################
+# Rules by tags #
+#################
+blank_lines: false             # Error on blank lines
```
tunetheweb, line 38: Missing newline at the end.

```suggestion
blank_lines: false             # Error on blank lines

```

Probably should enable `VALIDATE_YAML: true` too to catch this sort of thing. But it again needs a config file: https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/.github/linters/.yaml-lint.yml

.github/workflows/linter.yml
```diff
@@ -51,5 +51,7 @@ jobs:
         uses: github/super-linter@v4
         env:
           VALIDATE_ALL_CODEBASE: false
+          VALIDATE_JAVASCRIPT_ES: true
+          VALIDATE_MARKDOWN: true
           DEFAULT_BRANCH: main
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
tunetheweb, line 57: Missing newline at the end.

```suggestion
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}

```

dist/crawl_links.js
```diff
@@ -0,0 +1,40 @@
+const intersectRect = function (r1, r2) {
```
tunetheweb, line 1: Should we add a comment at the top of this file to explain what we're doing and why? A lot of info in the PR, but nothing in here and am sure we'll soon forget why this is here.

pmeenan:
@tunetheweb I added the newlines (and yaml config) but as far as I can tell, the HA yaml config files you linked to don't have empty newlines at the end either:

https://github.com/HTTPArchive/legacy.httparchive.org/blob/master/.github/linters/.markdown-lint.yml

https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/.github/linters/.yaml-lint.yml

tunetheweb:
That’s cause I never enabled YAML Lint in that. It was legacy so didn’t waste too much of my time with it. But good to do this one properly.

tunetheweb:
Oh hut Almanac is not legacy. Maybe YAML Lint does not check for a trailing newline? I just presumed it did.

tunetheweb (approved):

tunetheweb:
BTW, looks like YAMLLint DOES check for that https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.new_line_at_end_of_file

And there definitely is a trailing new inline in both those files in the other repo when I look at them in VSCode.

dist/crawl_links.js
```diff
@@ -0,0 +1,45 @@
+/*
+Collects same-origin in-viewport links from img, h[1-6] tags,
+sorted by size from largest to smallest.
+This is the seed for generating a list of URLs to crawl from the given page.
+*/
+const intersectRect = function (r1, r2) {
+    return !(r2.left > r1.right || 
+              r2.right < r1.left ||
```
tunetheweb, line 8:
```suggestion
    return !(r2.left > r1.right ||
              r2.right < r1.left ||
```

dist/crawl_links.js
```diff
@@ -0,0 +1,45 @@
...
+              r2.top > r1.bottom ||
+              r2.bottom < r1.top);
+};
+const sameOrigin = function(uri1, uri2){
+    uri1 = new URL(uri1);
+    uri2 = new URL(uri2);
+    if(uri1.host !== uri2.host) return false;
+    if(uri1.port !== uri2.port) return false;
+    if(uri1.protocol !== uri2.protocol) return false;
+    return true;
```
rviscomi, line 18: Can we simplify this?
```suggestion
    return uri1.origin == uri2.origin;
```

dist/crawl_links.js
```diff
@@ -0,0 +1,45 @@
...
+This is the seed for generating a list of URLs to crawl from the given page.
+*/
+const intersectRect = function (r1, r2) {
+    return !(r2.left > r1.right ||
+              r2.right < r1.left ||
+              r2.top > r1.bottom ||
+              r2.bottom < r1.top);
+};
+const sameOrigin = function(uri1, uri2){
+    uri1 = new URL(uri1);
```
rviscomi, line 13: The `link.href` may be relative, so we can make it absolute by passing `uri2` as the second arg. This means that the order of the arguments is not interchangeable so it might also make sense to name them more definitively like `anchorHref` and `documentHref`.
```suggestion
    uri1 = new URL(uri1, uri2);
```

dist/crawl_links.js
```diff
@@ -0,0 +1,45 @@
+/*
+Collects same-origin in-viewport links from img, h[1-6] tags,
+sorted by size from largest to smallest.
+This is the seed for generating a list of URLs to crawl from the given page.
```
rviscomi, line 4: Do we also need any logic in the custom metric itself to return an empty list if the given page is already a secondary test page? Or will that be handled by the test server?

pmeenan:
@rviscomi not sure why it's not letting me comment inline, but...

the .href property on the HTMLAnchorElement JS API is fully expanded (automatically), even for relative URLs in markup so we shouldn't have to do anything special when comparing.

As far as returning an empty list when it is a secondary page, I was going to handle that in the actual logic (in case we decide to go deeper on some ranks).  Basically keep the agent logic clean:

* Controller specifies a pub/sub queue to post "completed" tests to along with the metrics to include (the crawl_links custom metric)
* Controller includes the crawl depth in the metadata for the test
* When running a secondary page, it won't set the pub/sub "completed" queue at all since the controller won't have to do anything with the results and submit more tests

That way it will be an easy merge with the public WPT data and just use a generic capability for sending results to a pub/sub queue as configured on a per-test basis (the retry queue is already specified in the job so it's very similar).

pmeenan:
One last question - should we limit the number of URLs returned to make sure it doesn't blow up for some sites? Say, at most, 20 URLs after sorting?

pmeenan:
Changed the logic a bit. Instead of selectively looking at img and h tags, it now grabs all of the links on the page directly.
* Makes sure they are in the viewport
* Makes sure they are visible
* Aggregates each dest by the size of all of the links to it
* Limits the result to 20 URLs

This handles the edge cases of text-only sites like wikipedia and sites that use DIV's to structure their page and since it still uses the area covered by the link, large images or headings still get priority.

The visibility check keeps the crawl from extracting only carousel links so it can have a variety of link types from sites. Otherwise if they have a large carousel with lots of elements, they would dominate the ordered list.

dist/crawl_links.js
```diff
@@ -0,0 +1,53 @@
...
+for (let e of elements) {
+    try {
+        const rect = e.getBoundingClientRect();
+        const style = window.getComputedStyle(e);
+        if (rect.width > 1 && rect.height > 1 &&
+                e.offsetParent !== null &&
+                style.visibility !== 'hidden' &&
+                style.display !== 'none' &&
+                intersectRect(rect, viewport)) {
+            if (e && e.href != documentOrigin && sameOrigin(e.href, documentOrigin)) {
```
rviscomi, line 39: Could we also exclude links to the same page, like `#foo`?

I [tested](https://www.webpagetest.org/result/220418_AiDcZG_B85/1/details/#waterfall_view_step1) this on the NYT home page with mobile emulation and got these 5 results:

```
[
    "https://www.nytimes.com/live/2022/04/18/world/ukraine-russia-war-news",
    "https://www.nytimes.com/live/2022/04/18/sports/boston-marathon",
    "https://www.nytimes.com/subscription?campaignId=79FKF",
    "https://www.nytimes.com/#site-content",
    "https://www.nytimes.com/#site-index"
]
```

The last two should probably be omitted.

It also makes me wonder if we should handle mobile differently, since the viewport is much smaller. For example, should we consider nav items hidden behind the hamburger menu?

Desktop [results](https://www.webpagetest.org/result/220418_AiDcB8_B83/1/details/#waterfall_view_step1) for comparison:

```
[
    "https://www.nytimes.com/live/2022/04/18/world/ukraine-russia-war-news",
    "https://www.nytimes.com/2022/04/18/world/europe/ukraine-forces-cluster-munitions.html",
    "https://www.nytimes.com/2022/04/18/world/europe/russian-economy-bleak-assessments.html",
    "https://www.nytimes.com/live/2022/04/18/sports/boston-marathon",
    "https://www.nytimes.com/subscription?campaignId=79FKF",
    "https://www.nytimes.com/international/?action=click&region=Editions&pgtype=Homepage",
    "https://www.nytimes.com/section/t-magazine",
    "https://www.nytimes.com/section/realestate",
    "https://www.nytimes.com/es/",
    "https://www.nytimes.com/section/magazine",
    "https://www.nytimes.com/section/business",
    "https://www.nytimes.com/ca/?action=click&region=Editions&pgtype=Homepage",
    "https://www.nytimes.com/section/science",
    "https://www.nytimes.com/section/opinion",
    "https://www.nytimes.com/section/politics",
    "https://www.nytimes.com/section/health",
    "https://www.nytimes.com/section/sports",
    "https://www.nytimes.com/section/books",
    "https://www.nytimes.com/section/travel",
    "https://www.nytimes.com/section/world"
]
```

pmeenan:
@rviscomi 
Just added logic to strip the hash marks when comparing so it handles the NYT mobile case.

As far as viewport goes and the hamburger menu, I'm a bit torn.  I could append hidden URLs after the list of visible URLs to make sure we have decent coverage on mobile too but that will likely favor content links over hamburger menu links and at least on the NYT page, the collapsed hamburger menu items don't have an "in-viewport" position until they are expanded.

We could potentially do a second pass, collecting all non-viewport links ordered by size and append them to the list after the visible in-viewport ones, just to make sure we have enough to crawl several pages if we need them but that will likely favor stories vs the menu.

tunetheweb:
Could you merge `main` into this branch @pmeenan ? Just fixed the linter set up in #5

pmeenan:
Updated it to add the size-sorted non-visible URLs after the list of visible URLs. It doesn't pay attention to position so large stories late in the NYT feed may outrank earlier ones but in-viewport will always get priority.

dist/crawl_links.js
```diff
@@ -0,0 +1,65 @@
...
+    const documentOrigin = window.location.href;
+    const elements = document.links;
+    for (let e of elements) {
+        try {
+            const rect = e.getBoundingClientRect();
+            const style = window.getComputedStyle(e);
+            const is_visible = e.offsetParent !== null &&
+                               style.visibility !== 'hidden' &&
+                               style.display !== 'none' &&
+                               intersectRect(rect, viewport);
```
tunetheweb, line 38: Annoyingly the editorconfig linter doesn't handle hanging indents like this when non-even number of indent spaces.

We could turn the indent checking off, but it is useful in other cases.

```suggestion
            const is_visible = e.offsetParent !== null &&
                                style.visibility !== 'hidden' &&
                                style.display !== 'none' &&
                                intersectRect(rect, viewport);
```

rviscomi:
LGTM! One last comment. WDYT about attempting to deduplicate similar pages? For example, should we take steps to avoid testing all NYT article pages since they probably wouldn't tell us anything new? We could look for pages that have distinct top-level subdirectories, ie /live, /subscription, /2022, /interactive:

```
[
    "https://www.nytimes.com/live/2022/04/18/world/ukraine-russia-war-news",
    "https://www.nytimes.com/live/2022/04/18/sports/boston-marathon",
    "https://www.nytimes.com/subscription?campaignId=79FKF",
    "https://www.nytimes.com/2022/04/18/travel/south-georgia-island-recovery.html",
    "https://www.nytimes.com/2022/04/13/realestate/why-you-should-plant-a-garden-thats-wasp-friendly.html",
    "https://www.nytimes.com/2022/04/18/business/economy/jim-farley-ford-lightning.html",
    "https://www.nytimes.com/2022/04/17/business/china-economy-gdp.html",
    "https://www.nytimes.com/2022/04/18/world/europe/france-election-economy.html",
    "https://www.nytimes.com/2022/04/18/us/homeless-deaths-los-angeles.html",
    "https://www.nytimes.com/2022/04/17/business/disney-politics-florida.html",
    "https://www.nytimes.com/2022/04/18/opinion/sway-kara-swisher-tony-kushner.html",
    "https://www.nytimes.com/2022/04/17/nyregion/covid-questions-families.html",
    "https://www.nytimes.com/2022/04/18/opinion/elizabeth-warren-democrats-biden-midterms.html",
    "https://www.nytimes.com/2022/04/18/business/europe-green-energy-investors.html",
    "https://www.nytimes.com/2022/04/18/well/move/foot-health-pandemic.html",
    "https://www.nytimes.com/interactive/2021/11/12/well/move/back-stretches.html",
    "https://www.nytimes.com/2022/02/02/well/move/body-neutrality-exercise.html",
    "https://www.nytimes.com/interactive/2021/10/14/smarter-living/wirecutter/how-to-choose-running-shoes.html",
    "https://www.nytimes.com/2022/04/15/well/move/peloton-workouts-at-home.html",
    "https://www.nytimes.com/live/2022/04/18/world/covid-19-mandates-cases-vaccine"
]
```

pmeenan:
I get a bit nervous about trying to de-dupe based on relative paths because different sites will structure things differently. If we're going to do it, we should do it in the controller just to make sure we still have the largest links as the list to work off of.

We could do something where we make passes on the list, increasing the directory depth of uniqueness until we have enough URLs.

It's worth noting that the URL structure here looks like the /live/2022/* eventually become /2022/* so those will likely be the same template and then different months will register as different but still be article templates (and the /interactive/ looks like a 2021 format).

I think we're probably best-off if we just crawl a lot more pages for high-ranking sites and drop off as the ranking changes.

Top 1000: 15 additional pages
Top 10000: 5 additional pages
Top 100000: 2 additional pages
Remainder: 1 additional page

rviscomi:
Agreed that's probably best left to the controller. I also hadn't considered varying the number of secondary pages by rank, but that's a good idea.

One worry is that it might introduce more bias, so we'll need to be careful in our analyses. When we test these secondary pages, could we pass their index through to the HAR data? For example, we should be able to filter pages in BigQuery by only the home pages, only the most prominent secondary pages (those at index 0), and everything else.

That's not something we need to worry about for this PR so I'll LGTM and merge it if @tunetheweb is still happy.

rviscomi (approved):

tunetheweb:
I'm happy - merge away!

pmeenan:
Oh, absolutely.  I was expecting that we'd include both the crawl_depth and link_depth as metrics in the metadata. That way if we decide to crawl deeper (links from within the linked pages) we can filter by either.

I wasn't planning to enable the controller-side of things yet so we can get a regular crawl under our belts before worrying about leaf pages but I can add the depth metadata now since it won't change.

PR #4, opened 2022-04-15
Add editorconfig file and standardise custom metric files

tunetheweb:
Adds the `.editorconfig` file as used in the Almanac for those using [EditorConfig](https://editorconfig.org/) to help enforce file format styles.

Including `spaces` over `tabs` (let battle commence!)

To actually make use of this may require installing the `editorconfig` node module via npm (we suggest installing globally - `npm install -g editorconfig` ) and then installing the extension for your IDE (for example there is a [Visual Studio Code extension](https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig)).


dist/Colordepth.js
```diff
@@ -1,4 +1,4 @@
-var calcColorDepth = function() {
```
tunetheweb, line 1: Pro tip, review this with hiding whitespace:

image

pmeenan:
LGTM

pmeenan (approved):

PR #9, opened 2022-05-09
Switch to using $WPT_URL for the crawl links extraction

pmeenan:
The agents now support a string substitution where $WPT_URL will be replaced by the last URL that WPT navigated to as part of the test. This is needed to make sure that the links extracted are of the same origin as the page we initially navigated to, not necessarily wherever redirects took us.

### DIFF
--- a/dist/crawl_links.js
+++ b/dist/crawl_links.js
@@ -26,7 +26,7 @@ const viewport = {
 };
 const getLinks = function(visibleOnly){
     let links = {};
-    const documentOrigin = window.location.href;
+    const documentOrigin = $WPT_URL;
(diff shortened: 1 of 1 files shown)


rviscomi:
LGTM but confused by "last URL that WPT navigated to as part of the test". Is there ever more than 1 URL navigated? I assume redirects aren't considered navigations.

rviscomi (approved):

pmeenan:
@rviscomi If you have a multi-step script then there can be multiple navigations to different pages before the custom metric is run.  In the HA case, it is only ever 1 (at least for now).

PR #10, opened 2022-05-10
Detect fetchpriority attribute

rviscomi:
Detect the `fetchpriority` attribute in the `priority-hints` custom metric.

Test page: https://www.runnersworld.com/
Test results: https://webpagetest.httparchive.org/result/220510_B2_1/

### DIFF
--- a/dist/almanac.js
+++ b/dist/almanac.js
@@ -215,7 +215,7 @@ return JSON.stringify({
   'priority-hints': (() => {
     // Returns a JSON array of prioritized nodes and their key/value attributes.
     // Used by 19.8, 19.9, and 19.10.
-    var nodes = document.querySelectorAll('link[importance], img[importance], script[importance], iframe[importance]');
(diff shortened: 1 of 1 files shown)


dist/almanac.js
```diff
@@ -215,7 +215,7 @@ return JSON.stringify({
   'priority-hints': (() => {
     // Returns a JSON array of prioritized nodes and their key/value attributes.
     // Used by 19.8, 19.9, and 19.10.
-    var nodes = document.querySelectorAll('link[importance], img[importance], script[importance], iframe[importance]');
+    var nodes = document.querySelectorAll('link[importance], link[fetchpriority], img[importance], img[fetchpriority], script[importance], script[fetchpriority], iframe[importance], iframe[fetchpriority]');
```
pmeenan, line 218: We should remove the checks for importance and leave just fetchpriority. `importance` no longer works in Chrome and was never implemented anywhere else.

pmeenan (changes requested):

dist/almanac.js
rviscomi: It might still be useful to see instances where sites are using the deprecated attribute. We'll be able to distinguish between them in SQL.

pmeenan (approved):

pmeenan:
LGTM

PR #12, opened 2022-05-17
New Custom metric to check for valid head

csliva:
According to these [webmaster guidlines](https://developers.google.com/search/docs/advanced/guidelines/valid-html) a head will be terminated by invalid HTML elements within the head node. This is worth tracking because Googlebot will terminate the head tag early and causing potential SEO issues. 

This additional custom metric will return a boolean if any invalid elements are found.

**Tests**
URL with broken head: https://crawler-test.com/other/non_head_tag_in_head
WPT test URL: https://www.webpagetest.org/details.php?test=220518_AiDcAF_DFS&run=1&cached=0
Output: {"invalidElements":["div"],"invalidHead":true}

URL with correct head: https://developer.mozilla.org/en-US/
WPT test URL: https://www.webpagetest.org/details.php?test=220518_AiDcP0_DQW&run=1&cached=0
Output: {"invalidElements":[],"invalidHead":false}

URL with broken head: https://crawlgo.fly.dev/badhead
WPT test URL: https://www.webpagetest.org/details.php?test=220525_BiDcTF_FW7&run=1&cached=0
Output: {"invalidElements":["div", "p"],"invalidHead":true}


tunetheweb:
Hey @csliva could you run some test runs with this new custom metric on WebPageTest for a sample of common web pages (ideally one with a broken HEAD too that this detects) and include the links to the results as a comment in the PR?

More details here: https://github.com/HTTPArchive/custom-metrics#testing

dist/valid-head.js
```diff
@@ -0,0 +1,8 @@
+var valid_elements = ['TITLE', 'META', 'LINK', 'SCRIPT', 'STYLE', 'BASE', 'NOSCRIPT', 'TEMPLATE']
+var elems = document.head.getElementsByTagName("*");
+for (var i=0; i<elems.length;i++) {
+    if (!valid_elements.includes(elems[i].tagName)) {
+        return true;
```
rviscomi, line 5: It might be useful metadata to return the list of invalid elements for further analysis/exploration.

csliva:
> 

Good call @tunetheweb, tests failed. WPT moves invalid elements back into the body so I think I'll have to use $WPT_BODIES.

csliva:
**URL with broken head:** https://crawler-test.com/other/non_head_tag_in_head
**WPT test URL:**  https://www.webpagetest.org/details.php?test=220518_AiDcAF_DFS&run=1&cached=0
**Output:**  `{"invalidElements":["div"],"invalidHead":true}`


**URL with correct head:** https://developer.mozilla.org/en-US/
**WPT test URL:**  https://www.webpagetest.org/details.php?test=220518_AiDcP0_DQW&run=1&cached=0
**Output:**  `{"invalidElements":[],"invalidHead":false}`

dist/valid-head.js
```diff
@@ -0,0 +1,23 @@
+// Used for WPT
+// [valid-head]
+// Use DOMParser to extract head elements so browser can't truncate the head
+// Loop through each node and check against valid elements array
+// Returns object with ARRAY and BOOL
+var html = $WPT_BODIES[0].response_body;
+var parser = new DOMParser();
+var xmlDoc = parser.parseFromString(html,"text/xml");
```
rviscomi, line 8: Oh interesting. How different is this approach from `createHTMLDocument`? See the wpt_bodies custom metric for example:

https://github.com/HTTPArchive/custom-metrics/blob/4c6bc0493b1b51dc021d6dcf7471b80224dc9228/dist/wpt_bodies.js#L82-L84

Is parsing as XML somehow more strict? Or for the purposes of validating the head, is it actually necessary to parse as XML?

rviscomi, line 8: Looks like this might throw a `TypeError`. Could we add try/catch handling?

rviscomi:
@csliva ping, a couple of comments still outstanding

jroakes:
This looks really great!  One note, we may want to force lower-case on tag name matching.

dist/valid-head.js
csliva: Done! Sorry for the wait

csliva: So XML allows me to extract the head elements as they are sent. The createHTMLDocument truncates the document just like the browser would. I ran some additional tests and my function returns the number of element nodes correctly. However, once parsed as XML, it is NOT valid HTML anymore.

csliva:
Additional tests with a test site I built to make sure XML parsing worked. 
URL with broken head: https://crawlgo.fly.dev/badhead
WPT test URL: https://www.webpagetest.org/details.php?test=220525_BiDcTF_FW7&run=1&cached=0
Output: {"invalidElements":["div", "p"],"invalidHead":true}

dist/valid-head.js
```diff
@@ -0,0 +1,36 @@
...
+        for (var i=0; i<nodes.length;i++) {
+        let tagname = nodes[i].tagName.toLowerCase()
+        if (!valid_elements.includes(tagname)) {
+            data.invalidElements.push(tagname);
+            data.invalidHead = true;
+        };
+    }
+    return data;
+} catch (err) {
+    console.log(err);
```
rviscomi, line 32:
```suggestion
```

dist/valid-head.js
```diff
@@ -0,0 +1,36 @@
...
+  {
```
rviscomi, line 33: Are these extra braces needed?

dist/valid-head.js
```diff
@@ -0,0 +1,36 @@
...
+
+    [...head.getElementsByTagName("interrupt")].forEach(el => el.remove());
+    var nodes = head.getElementsByTagName("*");
+
+    let data = {
+        "invalidElements":  [],
+        "invalidHead": false
+    }
+    var valid_elements = ['title', 'meta', 'link', 'script', 'style', 'base', 'noscript', 'template']
+        for (var i=0; i<nodes.length;i++) {
```
rviscomi, line 23:
```suggestion
    for (var i = 0; i < nodes.length; i++) {
```

dist/valid-head.js
```diff
@@ -0,0 +1,36 @@
...
+    var html = $WPT_BODIES[0].response_body;
+    // create in interrupt element which prevents truncation.
+    html = html.replace('</head>', "<interrupt />\n</head>");
+
+    var doc = document.implementation.createHTMLDocument("New Document");
+    var head = doc.createElement("div");
+    head.innerHTML = html;
+
+    [...head.getElementsByTagName("interrupt")].forEach(el => el.remove());
+    var nodes = head.getElementsByTagName("*");
```
rviscomi, line 16: Not quite following this part. You're creating a `div` element, setting its contents to be the entire response body, removing the `interrupt` element, and querying everything else. How does that entire that you're only querying elements in the `head`?

csliva: We create an `interrupt` element just before the head closes. The rest of the HTML document ends up in the `interrupt` element and then we delete everything inside of that. 
It's like pulling a reverse uno card on the HTML trying to correct itself.

jroakes: Here is a demo showing this working on Chrome.  https://www.screencast.com/t/q12dpBfUaqy  For full disclosure, I am not 100% sure how this works, but noticed it a while ago while working on something else.

rviscomi: 🤷🪄

rviscomi (approved):

dist/valid-head.js
```diff
@@ -0,0 +1,35 @@
...
+        if (!valid_elements.includes(tagname)) {
+            data.invalidElements.push(tagname);
+            data.invalidHead = true;
+        };
+    }
+    return data;
+} catch (err) {
+  {
+      return {error: err.toString()};
+  }
```
rviscomi, line 34:
```suggestion
    return {error: err.toString()};
```

dist/valid-head.js
rviscomi: Querying these results now 👀 

invalid | pages | n
-- | -- | --
img | 1,470,348 | 7,022,280
div | 652,648 | 21,726,625
a | 342,369 | 21,075,376
p | 212,880 | 6,538,475
span | 208,986 | 9,677,474
br | 176,658 | 5,512,635
iframe | 171,274 | 231,499
input | 159,036 | 1,800,872
ul | 141,196 | 1,683,013
li | 141,182 | 10,075,377
form | 137,864 | 401,917
h1 | 127,850 | 429,224
button | 119,402 | 533,072
h2 | 117,586 | 756,110
h3 | 103,785 | 1,045,688
svg | 101,265 | 1,106,651
footer | 97,547 | 424,299
header | 96,379 | 377,136
path | 95,030 | 1,681,346
label | 94,958 | 570,545
nav | 93,122 | 177,827
b | 92,002 | 1,774,036
strong | 83,602 | 1,734,224
table | 78,495 | 883,732
tbody | 78,239 | 879,676
tr | 78,226 | 2,687,622
i | 77,575 | 1,264,969
td | 77,514 | 5,867,209
article | 67,777 | 555,650
g | 67,408 | 867,298

PR #13, opened 2022-05-19
Script injection and UA custom metric

rviscomi:
Add a new script injection directory and seed it with an observer script that detects `navigator.userAgent` usage.

Add a new property to the `privacy` custom metric to output the value of the UA observer.

Tests:

- https://webpagetest.httparchive.org/result/220519_0T_D/1/details/#waterfall_view_step1


dist/privacy.js
```diff
@@ -311,4 +311,6 @@ return JSON.stringify({
       'navigator.+geolocation.+watchPosition'
     ),
   },
+  
+  userAgent: observers.userAgent
```
pmeenan, line 315: Should this be `httparchive_observers`?

pmeenan (changes requested):

dist/privacy.js
rviscomi: Fixed

pmeenan (approved):

pmeenan:
LGTM modulo the lint issues

Issue #16, opened 2022-05-23
Performance 2022 custom metrics


rviscomi:
Progress on https://github.com/HTTPArchive/almanac.httparchive.org/issues/2890
## Antipatterns
- [x] LCP image source dynamically changed

Add a new custom metric to performance.js that detects whether the LCP image (if applicable) had its image source dynamically changed.

For example, if the raw HTML source was served as `<img data-src="cat.jpg">` but the fully-rendered LCP image was `<img src="cat.jpg">`, some kind of lazy-loading library seems to be at play.

## Gaming the Metrics
- [ ] Filtering for synthetic tests -  Add a new custom metric to performance.js that detects if there is any code in HTML response to filter some resources based on synthetic tests user agent such as lighthouse/GTmatrix. 
- [ ] LCP Animation & Overlay Hack - Detect if LCP element is 'hidden' by opacity animation. 
- [ ] LCP Svg Overlay Hack: Detect image tag at top of the body containing transparent SVG at 99vh & 99vw, with the `pointer-events: none` CSS style.
- [ ] CLS Animation Hack - Detect if layout shift element is 'hidden' by opacity animation
- [ ] FID iframe Hack - Detect if viewport is covered in transparent iframe -> after pageload event fires, iframe is removed. Will require adding a script to run before the page loads and comparing the initial response body with the DOM after page load (see [Rick's livestream](https://www.youtube.com/watch?v=71lA-p15NhI) for tutorial)
- [ ] [Nice to Have] LCP Element Render Delay - Number of sync resources (ResponseEnd time of resources (timestamp) < LCP resource startTime),  for LCP elements like H1 /any HTML tag, based on element HTML position

Note: The LCP Animation & CLS Animation metrics may be able to be combined. What if we scan for all elements that have this animation applied, and return something like this?
```{ numberOfElements: 6, isAppliedToLCPElement: true}```

cc @mel-ada @konfirmed @25prathamesh @silent1mezzo

rviscomi:
In general that'd be really helpful to know that there was some delay in fetching. Would we be able to tell why fetching was delayed without further instrumentation like this?

pmeenan:
You may be able to get some sense of "why" by looking at the initiator data to see if it was triggered by JS but I'm not sure how well the initiator catches all of the cases.

rviscomi:
[...]

Issue #15, opened 2022-05-23
Investigate impact of first response body not being HTML


rviscomi:
Some custom metrics rely on `$WPT_BODIES[0]` being the main HTML document. However, we've seen some edge cases (on up to 15% of pages) where the first request does not correspond to the main document. These custom metrics would assumedly be processing the data for the wrong response body.

Investigate whether this is actually happening and how to fix it, if so.

rviscomi:
It doesn't seem like the `$WPT_BODIES` object is affected by this issue.

[Here](https://webpagetest.httparchive.org/result/220402_Dx90_HKKF/1/details/#waterfall_view_step1) is a test from the 2022_04_01_desktop crawl in which the URL was incorrectly parsed as `http://cacerts.digitalcertvalidation.com/TrustAsiaTLSRSACA.crt`. The request at `entries[0]` in the HAR corresponds to the certificate and `_is_base_page` is set to true.

```json
{
    "_full_url": "http://cacerts.digitalcertvalidation.com/TrustAsiaTLSRSACA.crt",
    "_is_base_page": true,
    "_index": 0
}
```

I [reran](https://webpagetest.httparchive.org/result/220526_5P_9/1/details/#waterfall_view_step1) the test with a custom metric that outputs the full `$WPT_BODIES` object. The object schema is not quite the same as the HAR but it's clear that the first item is the HTML document itself, not the cert:

```json
{
    "url": "https://52.mk/",
    "type": "Document"
}
```

So I think we're ok.

pmeenan:
The waterfall and main request data are post-processed using the netlog trace events.  The $WPT_BODIES (and $WPT_REQUESTS) use the dev tools request details which don't see things like OCSP checks so the first request should (hopefully) always be the actual navigation.

PR #20, opened 2022-05-24
Performance 2022 custom metrics

rviscomi:
Progress on #16 

Additional metadata about the LCP element:

- `is_lcp_discoverable`: boolean whether it was discoverable in the raw HTML
- `is_lcp_preloaded`: boolean whether it was preloaded
- `boundingClientRect`: dimensions and position of the rendered element
- `naturalHeight`, `naturalWidth`: intrinsic image dimensions
- `lcp_resource`: full WPT response metadata
- `raw_lcp_element`: node name and attributes of the element in the raw HTML
- `styles`: allowlist of computed style prop/value pairs, set to `background-image` for now

Everything is applicable to images. The `boundingClientRect` will also be available for text-based LCP elements.

## Tests

- Image preloaded image, not discoverable
  - URL: https://rviscomi.github.io/demos/lcp_preloaded.html
  - Results: https://webpagetest.httparchive.org/result/220526_QT_12/1/details/#waterfall_view_step1
- Image not preloaded, discoverable
  - URL: https://www.nytimes.com/
  - Results: https://webpagetest.httparchive.org/result/220526_ND_10/1/details/#waterfall_view_step1
- Sourceset image not preloaded, discoverable
  - URL: https://almanac.httparchive.org/en/2021/css
  - Results: https://webpagetest.httparchive.org/result/220526_EE_14/1/details/#waterfall_view_step1
- Background image not preloaded, not discoverable?
  - URL: https://codepen.io/rviscomi/pen/wvyPQoM
  - Results: https://webpagetest.httparchive.org/result/220526_H2_15/1/details/#waterfall_view_step1
- Not an image
  - URL: https://www.example.com/
  - Results: https://webpagetest.httparchive.org/result/220526_V1_11/1/details/#waterfall_view_step1

### DIFF
--- a/dist/performance.js
+++ b/dist/performance.js
@@ -23,22 +23,58 @@ function getLcpElement() {
             resolve(naiveLcpEntry);
         }).observe({ type: "largest-contentful-paint", buffered: true });
     }).then(({ startTime, element, url, size, loadTime, renderTime }) => {
-        let attributes = [];
-        for (let index = 0; index < element.attributes.length; index++) {
-            const ele = element.attributes.item(index);
-            attributes[index] = { name: ele.name, value: ele.value };
-        }
-
+        const attributes = getAttributes(element);
+        const styles = getComputedStyles(element, ['background-image']);
         return {
             startTime,
-            nodeName: element.nodeName,
+            nodeName: element?.nodeName,
             url,
             size,
             loadTime,
             renderTime,
             attributes,
+            boundingClientRect: element?.getBoundingClientRect().toJSON(),
+            naturalWidth: element?.naturalWidth,
+            naturalHeight: element?.naturalHeight,
+            styles
         };
     });
+}
+
+function getAttributes(element) {
+    if (!element) {
+        return null;
+    }
+
+    return Array.from(element.attributes).map(attr => {
+        return {
+            name: attr.name,
+            value: attr.value
+        };
+    });
+}
+
+function getComputedStyles(element, properties) {
+    if (!element) {
+        return null;
+    }
+
+    const styles = getComputedStyle(element);
+    return Object.fromEntries(properties.map(prop => ([prop, styles.getPropertyValue(prop)])));
+}
+
+function summarizeLcpElement(element) {
+    if (!element) {
+        return null;
+    }
+
+    const nodeName = element.nodeName;
+    const attributes = getAttributes(element);
+
+    return {
+        nodeName,
+        attributes
+    }
 }
 
 function getWebVitalsJS() {
@@ -48,18 +84,24 @@ function getWebVitalsJS() {
     }).map(har => har.url);
 }
 
-return Promise.all([getLcpElement()]).then(lcp_elem_stats => {
-    const lcpUrl = lcp_elem_stats[0].url;
+return Promise.all([getLcpElement()]).then(([lcp_elem_stats]) => {
+    const lcpUrl = lcp_elem_stats.url;
     const rawDoc = getRawHtmlDocument();
     let isLcpDiscoverable = null;
     let isLcpPreloaded = null;
     let responseObject = null;
+    let rawLcpElement = null;
     const isLcpExternalResource = lcpUrl != '';
     if (isLcpExternalResource) {
-        isLcpDiscoverable = !!Array.from(rawDoc.querySelectorAll('picture source, img')).find(i => {
-            const src = i.src || i.srcset;
+        rawLcpElement = Array.from(rawDoc.querySelectorAll('picture source, img')).find(i => {
+            let src = i.src;
+            if (i.nodeName == 'SOURCE') {
+                src = new URL(i.srcset, location.href).href;
+            }
+
             return src == lcpUrl;
         });
+        isLcpDiscoverable = !!rawLcpElement;
         isLcpPreloaded = !!Array.from(rawDoc.querySelectorAll('head link')).find(link => {
             return link.rel == 'preload' && link.href == lcpUrl;
         });
@@ -74,9 +116,12 @@ return Promise.all([getLcpElement()]).then(lcp_elem_stats => {
 
     return {
         lcp_elem_stats,
+        raw_lcp_element: summarizeLcpElement(rawLcpElement),
         lcp_resource: responseObject,
         is_lcp_discoverable: isLcpDiscoverable,
         is_lcp_preloaded: isLcpPreloaded,
         web_vitals_js: getWebVitalsJS()
     };
+}).catch(error => {
+    return {error};
 });
--- a/dist/performance.js
+++ b/dist/performance.js
@@ -93,6 +93,7 @@ return Promise.all([getLcpElement()]).then(([lcp_elem_stats]) => {
     let rawLcpElement = null;
     const isLcpExternalResource = lcpUrl != '';
     if (isLcpExternalResource) {
+        // Check if LCP resource reference is in the raw HTML (as opposed to being injected later by JS)
         rawLcpElement = Array.from(rawDoc.querySelectorAll('picture source, img')).find(i => {
             let src = i.src;
             if (i.nodeName == 'SOURCE') {
@@ -102,7 +103,7 @@ return Promise.all([getLcpElement()]).then(([lcp_elem_stats]) => {
             return src == lcpUrl;
         });
         isLcpDiscoverable = !!rawLcpElement;
-        isLcpPreloaded = !!Array.from(rawDoc.querySelectorAll('head link')).find(link => {
+        isLcpPreloaded = !!Array.from(rawDoc.querySelectorAll('link')).find(link => {
             return link.rel == 'preload' && link.href == lcpUrl;
         });
         responseObject = response_bodies.find(r => {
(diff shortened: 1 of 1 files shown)


rviscomi:
> Background image not preloaded, not discoverable?

[This](https://webpagetest.httparchive.org/result/220526_H2_15/1/details/#waterfall_view_step1) is an interesting case. The metric evaluates to not discoverable because there is no src or srcset on the page set to that image URL. But `lcp_resource.initiator.type` is `"parser"` so it seems to be discoverable? That might only be the case because the CSS is in an inline style block, haven't tested it with an external stylesheet. Should we drop the discoverable field and always check the initiator type?

dist/performance.js
```diff
@@ -11,34 +23,105 @@ function getLcpElement() {
...
+    const isLcpExternalResource = lcpUrl != '';
+    if (isLcpExternalResource) {
+        rawLcpElement = Array.from(rawDoc.querySelectorAll('picture source, img')).find(i => {
+            let src = i.src;
+            if (i.nodeName == 'SOURCE') {
+                src = new URL(i.srcset, location.href).href;
+            }
+
+            return src == lcpUrl;
+        });
```
tunetheweb, line 103: This confused me at first. But now 

```suggestion
    // Check if LCP resource reference is in the raw HTML (as opposed to being injected later by JS)
    if (isLcpExternalResource) {
        rawLcpElement = Array.from(rawDoc.querySelectorAll('picture source, img')).find(i => {
            let src = i.src;
            if (i.nodeName == 'SOURCE') {
                src = new URL(i.srcset, location.href).href;
            }

            return src == lcpUrl;
        });
```

dist/performance.js
```diff
@@ -11,34 +23,105 @@ function getLcpElement() {
...
+        isLcpDiscoverable = !!rawLcpElement;
+        isLcpPreloaded = !!Array.from(rawDoc.querySelectorAll('head link')).find(link => {
```
tunetheweb, line 105: I know best practice is to preload from the `<head>` but does it need it be in there? Any reason not to just look at all `link` elements?

dist/performance.js
```diff
@@ -11,34 +23,105 @@ function getLcpElement() {
...
         }).observe({ type: "largest-contentful-paint", buffered: true });
     }).then(({ startTime, element, url, size, loadTime, renderTime }) => {
-        let attributes = [];
-        for (let index = 0; index < element.attributes.length; index++) {
-            const ele = element.attributes.item(index);
-            attributes[index] = { name: ele.name, value: ele.value };
-        }
-
+        const attributes = getAttributes(element);
+        const styles = getComputedStyles(element, ['background-image']);
```
tunetheweb, line 27: Are we likely to want any more than this? Or should we just called this background-image?

```suggestion
        const backgroundImage = getComputedStyles(element, ['background-image']);
```

dist/performance.js
```diff
@@ -11,34 +23,105 @@ function getLcpElement() {
...
+            nodeName: element?.nodeName,
             url,
             size,
             loadTime,
             renderTime,
             attributes,
+            boundingClientRect: element?.getBoundingClientRect().toJSON(),
+            naturalWidth: element?.naturalWidth,
+            naturalHeight: element?.naturalHeight,
+            styles
```
tunetheweb, line 39:
```suggestion
            backgroundImage
```

tunetheweb (approved):
LGTM

A few comments from me (often for my benefit rather than yours!) but nothing to stop merging.

dist/performance.js
```diff
@@ -2,6 +2,18 @@
...
 const response_bodies = $WPT_BODIES;
 
+function getRawHtmlDocument() {
+    let rawHtml;
+    if (response_bodies.length > 0) {
+        rawHtml = response_bodies[0].response_body;
+    }
+
+    rawHtmlDocument = document.implementation.createHTMLDocument("New Document");
+    rawHtmlDocument.documentElement.innerHTML = rawHtml;
```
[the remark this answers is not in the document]
rviscomi: No idea, I copied this method from wpt_bodies for consistency. 😁 

Now that we have a dedicated custom-metrics repo, I'd love to use ES modules and do more code reuse. That would make it easier to have a common util for document creation in whichever strategy makes sense.

dist/performance.js
```diff
@@ -11,34 +23,105 @@ function getLcpElement() {
...
-            const ele = element.attributes.item(index);
-            attributes[index] = { name: ele.name, value: ele.value };
-        }
-
+        const attributes = getAttributes(element);
+        const styles = getComputedStyles(element, ['background-image']);
         return {
             startTime,
-            nodeName: element.nodeName,
+            nodeName: element?.nodeName,
```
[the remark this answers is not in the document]
rviscomi: `element` may be undefined sometimes (see the NYT test) so this ensures we only access the property if it exists.

More info on this syntax: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

dist/performance.js
```diff
@@ -11,34 +23,105 @@ function getLcpElement() {
...
 
-return Promise.all([getLcpElement()]).then(lcp_elem_stats => {
+return Promise.all([getLcpElement()]).then(([lcp_elem_stats]) => {
+    const lcpUrl = lcp_elem_stats.url;
+    const rawDoc = getRawHtmlDocument();
+    let isLcpDiscoverable = null;
+    let isLcpPreloaded = null;
+    let responseObject = null;
+    let rawLcpElement = null;
+    const isLcpExternalResource = lcpUrl != '';
```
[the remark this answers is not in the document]
rviscomi: Same, I use it to mean not inline.

dist/performance.js
rviscomi: Yeah I think it's likely that this will be used more.

dist/performance.js
rviscomi: Keeping general purpose for now

dist/performance.js
rviscomi: Good call, looks like `preload` is "body-ok" according to the [spec](https://html.spec.whatwg.org/multipage/links.html#linkTypes)

dist/performance.js
rviscomi: Added the comment inside the if statement

dist/performance.js
```diff
@@ -2,6 +2,18 @@
 
 const response_bodies = $WPT_BODIES;
 
+function getRawHtmlDocument() {
+    let rawHtml;
+    if (response_bodies.length > 0) {
+        rawHtml = response_bodies[0].response_body;
```
mel-ada, line 8: Do we have a confidence level for the 0th response body always being the raw HTML? I remember in the livestream you mentioned it's pretty likely - is there a way to test?

dist/performance.js
```diff
@@ -11,34 +23,106 @@ function getLcpElement() {
...
-            nodeName: element.nodeName,
+            nodeName: element?.nodeName,
             url,
             size,
             loadTime,
             renderTime,
             attributes,
+            boundingClientRect: element?.getBoundingClientRect().toJSON(),
+            naturalWidth: element?.naturalWidth,
+            naturalHeight: element?.naturalHeight,
```
mel-ada, line 38: Are we preemptively adding these to detect gaming the metrics hacks?

dist/performance.js
```diff
@@ -11,34 +23,106 @@ function getLcpElement() {
...
+            styles
```
mel-ada, line 39: To check my understanding - here we are creating a way to get the URL when an LCP image is added via the `background-image` CSS rule?

dist/performance.js
```diff
@@ -11,34 +23,106 @@ function getLcpElement() {
...
+            // Don't write the response body to custom metrics.
+            responseObject.response_body = undefined;
+        }
+    }
+
     return {
         lcp_elem_stats,
+        raw_lcp_element: summarizeLcpElement(rawLcpElement),
+        lcp_resource: responseObject,
+        is_lcp_discoverable: isLcpDiscoverable,
```
mel-ada, line 122: Thinking about naming conventions - if I came across this in a database a few years from now I might mistake this for 'Is there an LCP element at all?' instead of 'Is there an LCP element in the raw HTML?'. That said, I can't really think of a better name...does `is_lcp_dynamically_loaded` have less room for confusion?

mel-ada (approved):
LGTM - Non blocking comments mostly to check understanding

dist/performance.js
rviscomi: Good catch, yes we [looked into it](https://github.com/HTTPArchive/custom-metrics/issues/15) and we can rely on the 0th element to be the document. 👍

dist/performance.js
rviscomi: I was thinking this could be useful for analyzing oversized LCP images

dist/performance.js
rviscomi: Yes! The `url` field above will be the LCP image even for DIVs with background image styles, but in those cases we don't necessarily know that it was that background image that was the LCP. This helps to clarify that.

dist/performance.js
rviscomi: I like that. How about `is_lcp_statically_discoverable` or `is_lcp_speculatively_discoverable`?

mel-ada: Sounds good - if I had to pick I would vote for `statically`!

rviscomi:
Yeah hopefully that gives us enough metadata to figure it out later in SQL if there were any false negatives. We can analyze the next crawl to see how reliable the `discoverable` field is vs `initiator.type`.

PR #14, opened 2022-05-22
General purpose observer injection

rviscomi:
Changes the observer behavior to be able to handle wildcards in the property pathname. For example, `navigator.*` will observe all properties on the `navigator` object. Special handling was added for `__proto__` properties. Also adds the ability to capture the call stack for any particular properties.

### example.com

[Results](https://webpagetest.httparchive.org/result/220525_19_1A/1/details/#waterfall_view_step1)

It seems like the reset isn't being applied early enough. This might fix itself after the change is merged and the WPT agents consistently execute them in alphabetical order.

As expected, `navigator.userAgent` is not detected and the call stack is empty.

```json
{
    "call_stacks": {
        "navigator.userAgent": []
    },
    "navigator.clipboard": 2,
    "navigator.storage": 2,
    "navigator.serviceWorker": 4,
    "performance.clearMarks": 1,
    "performance.clearMeasures": 1,
    "performance.getEntriesByType": 9,
    "performance.timing": 9,
    "performance.timing.domInteractive": 1,
    "performance.timing.domContentLoadedEventStart": 1,
    "performance.timing.domContentLoadedEventEnd": 1,
    "performance.timing.domComplete": 1,
    "performance.timing.loadEventStart": 1,
    "performance.timing.loadEventEnd": 1,
    "Array.prototype.constructor": 57,
    "Array.prototype.push": 9,
    "Array.prototype.slice": 1,
    "Array.prototype.sort": 2,
    "Array.prototype.includes": 16,
    "Array.prototype.indexOf": 182,
    "Array.prototype.join": 1,
    "Array.prototype.forEach": 8,
    "Array.prototype.filter": 35,
    "Array.prototype.map": 21,
    "Array.prototype.some": 2,
    "Array.prototype.reduce": 6,
    "String.prototype.includes": 13,
    "String.prototype.indexOf": 83,
    "String.prototype.match": 50,
    "String.prototype.replace": 13,
    "String.prototype.slice": 1,
    "String.prototype.split": 12,
    "String.prototype.substring": 20,
    "String.prototype.trim": 11,
    "String.prototype.toLowerCase": 35
}
```

### https://codepen.io/rviscomi/pen/rNJMmgy

```js
const _hjs2 = navigator;
const __483 = 'userAgent'

function functionA() {
  console.log(_hjs2[__483])
}

function functionB() {
  console.log(_hjs2['userAgent'])
  console.log(navigator.userAgent)
}

functionA();
functionB();
```

[...]

PR #21, opened 2022-05-25
Privacy 2022 custom metrics

max-ostapenko:
adjustments:
- [ ] 

todo:
- [ ] add observer approach to check usage of privacy related methods and properties
- [ ] extend the list with more things tracked by fingerprinting
- [ ] check https://github.com/openwpm/OpenWPM approach for inspiration
- [ ] update wappalyzer libraries list
- [ ] updated compliance frameworks checks

### DIFF
--- a/dist/privacy.js
+++ b/dist/privacy.js
@@ -193,7 +193,7 @@ return JSON.stringify({
    *
    * Test site: https://www.theverge.com/
    */
-  navigator_doNotTrack: testPropertyStringInResponseBodies('navigator.+doNotTrack'),
+  navigator_doNotTrack: httparchive_observers.doNotTrack,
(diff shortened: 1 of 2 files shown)


rviscomi:
#14 now includes all of the `navigator` properties you've added here plus `document.featurePolicy` so after that PR is merged we could close this one as obsolete.

rviscomi:
#14 now includes all of the `navigator` properties you've added here plus `document.featurePolicy` so after that PR is merged we could close this one as obsolete.

rviscomi:
I think we can close this now that #14 is merged, all of the functionality should be built into the injected script and the data you need will be in the observers.js custom metric.

PR #22, opened 2022-05-25
Add two .well-known custom metrics for the security chapter

vikvanderlinden:
The `/change-password/` and `/resource-that-should-not-exist-whose-status-code-should-not-be-200/` have been added to the well-known metrics for the security chapter.

Tests:
https://webpagetest.org/result/220525_AiDc91_D7M/
https://webpagetest.org/result/220525_AiDcEC_D80/
https://webpagetest.org/result/220525_BiDc9Y_BXN/

### DIFF
--- a/dist/well-known.js
+++ b/dist/well-known.js
@@ -52,7 +52,7 @@ function parseResponseWithRedirects(url, parser) {
   return fetchWithTimeout(url)
     .then(request => {
       let resultObj = {};
-      resultObj['found'] = true;
+
       if (parser) {
         let promise = parser(request);
         if (promise) {
--- a/dist/well-known.js
+++ b/dist/well-known.js
@@ -48,6 +48,35 @@ function parseResponse(url, parser) {
   });
 }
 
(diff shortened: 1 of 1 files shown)


vikvanderlinden:
@GJFR @VictorLeP

SaptakS:
I think the found key is a bit confusing in these cases, since we don't expect to find a file or content in these scenarios unlike for `.well-known/security.txt`. `https://www.example.com/.well-known/resource-that-should-not-exist-whose-status-code-should-not-be-200/` is always supposed to return 404, which it does. It's more like a sanity check to see the servers are set up properly with status codes.

https://w3c.github.io/webappsec-change-password-url/response-code-reliability.html this has more on the actual specification.

vikvanderlinden:
I was basically skipping the check for status 200 and redirects in the new function `parseResponseWithRedirects` but I left the found key in. I removed it now because it did indeed not make much sense as @SaptakS suggested.

rviscomi (approved):

PR #25, opened 2022-05-25
"Reset" custom metric

rviscomi:
For stopping global observations, and to avoid observing our own custom metric work.


pmeenan (approved):
LGTM.

Would it be useful to have a `return true;` or similar to have the "00_reset" metric have a value that signals that it was run (maybe with a check to make sure it was enabled before running)?

rviscomi:
Can't think of a need for that type of output at the moment. Happy to expand the metric if those use cases come up later.

PR #26, opened 2022-05-25
Added longTasks to JavaScript custom metric using PerformanceObserver

kevinfarrugia:
I added a new PerformanceObserver to `observers.js` that populates a global array `httparchive_longTasks` (which I understand is injected into the page after the document has started loading).

My first attempt was to use promises inside the custom metric, but it wasn't returning any value.

Test: https://webpagetest.httparchive.org/result/220525_8A_2B/1/details/#waterfall_view_step1


rviscomi:
This might not be needed. WPT already includes long task info in the [HAR](https://webpagetest.httparchive.org/jsonResult.php?test=220525_8A_2B&pretty=1) (search `longTasks`).

kevinfarrugia:
@rviscomi I'm getting different results between the `_longTasks` included by default and my custom metric.

New example: https://webpagetest.httparchive.org/result/220526_YR_1F/1/details/#waterfall_view_step1

```
{"longTasks":{"httparchive_longTasks":[117,55,83,145,99,52,85,812,75],"count":9,"total":1523,"max":812}}
```

vs
```
_longTasks: [
  [379, 448],           // 69
  [809, 864],           // 55
  [2467, 2552],         // 85
  [2587, 2733],         // 146
  [3018, 3105],         // 87
];
```

This uses the injected script:

```
let httparchive_longTasks = [];
new PerformanceObserver(function(list) {
    let entries = list.getEntries();
    for (let i = 0; i < entries.length; i++) {
        httparchive_longTasks = [...httparchive_longTasks, entries[i].duration];
        console.log(JSON.stringify(entries[i].duration));
        console.log(httparchive_longTasks);
    }
}).observe({type: "longtask", buffered: true});
```

This allows me to check the `_consoleLog` in the HAR file to double-check:

```
"[117,55,83,145,99,52,85]"
```

52,85 seem to be genuine long tasks that should be included, while my guess is that the `812,75` are coming from our instrumentation.

Not sure if it is worth looking into.

rviscomi:
As long as you're using the `buffered` flag, I think you should be able to move this whole script into the custom metric and avoid the injected script, similar to how the [`performance.js`](https://github.com/HTTPArchive/custom-metrics/blob/b5e5d24ce4f06535e5f222ba9f2c9ae9923a8696/dist/performance.js#L7-L12) metric buffers LCP candidates.

If any of our custom metrics are creating the long tasks, you could run the observer early in the metric list by giving it a name like 01_long_tasks. Kind of hacky but I did something similar for 00_reset which must be executed first.

pmeenan:
_longTasks uses the Chrome trace data directly and looks at the duration of any main-thread tasks. The core of the logic is [here](https://github.com/WPO-Foundation/wptagent/blob/master/internal/support/trace_parser.py#L768)

kevinfarrugia:
Perfect, thank you @pmeenan. I will close this for the time being and we'll query on `_longTasks` directly.

PR #30, opened 2022-05-27
Add impact of sizes errors metrics

eeeps:
Progress on https://github.com/HTTPArchive/almanac.httparchive.org/issues/2884

As discussed in Slack starting around https://httparchive.slack.com/archives/C017HPKG614/p1652902287349589

I have successfully tested this on a couple of test pages, using WPT. Will run it against a handful of actual pages tomorrow; requesting some external code reviews from people familiar with how Chrome's selection algorithm works right now.

### DIFF
--- a/dist/responsive_images.js
+++ b/dist/responsive_images.js
@@ -485,8 +485,8 @@ function getImgData( img ) {
   // if there was a w descriptor -- was it accurate?
   if ( imgData.currentSrcWDescriptor &&
       imgData.approximateResourceWidth ) {
-    imgData.wDesciptorAbsoluteError = imgData.currentSrcWDescriptor - imgData.approximateResourceWidth;
-    imgData.wDesciptorRelativeError = imgData.wDesciptorAbsoluteError / imgData.approximateResourceWidth;
+    imgData.wDescriptorAbsoluteError = imgData.currentSrcWDescriptor - imgData.approximateResourceWidth;
+    imgData.wDescriptorRelativeError = imgData.wDescriptorAbsoluteError / imgData.approximateResourceWidth;
   }
 
   // get the server-reported mime type of the image
--- a/dist/responsive_images.js
+++ b/dist/responsive_images.js
@@ -522,7 +522,7 @@ function getImgData( img ) {
   }
 
   const smallestResourceThatsLargerThanDPR = function( annotatedCandidates ) {
-    if ( !annotatedCandidates || !annotatedCandidates.length || annotatedCandidates.length === 0 ) {
+    if ( !annotatedCandidates || !annotatedCandidates.length ) {
       return null;
     }
     const larger = annotatedCandidates.filter( c => c.linearDistance >= 0 );
@@ -537,7 +537,7 @@ function getImgData( img ) {
   }
 
   const resourceThatsClosestToDPR = function( annotatedCandidates ) {
-    if ( !annotatedCandidates || !annotatedCandidates.length || annotatedCandidates.length === 0 ) {
+    if ( !annotatedCandidates || !annotatedCandidates.length ) {
       return null;
     }
     return annotatedCandidates.sort( ( a, b ) => {
--- a/dist/responsive_images.js
+++ b/dist/responsive_images.js
@@ -499,6 +499,88 @@ function getImgData( img ) {
     Object.values( imgData.intrinsicOrExtrinsicSizing ).includes( 'intrinsic' )
   );
 
+  // what was the impact of incorrect sizes values?
+
+  // helpers...
+  function geometricDistance(a, b) {
+    if (a > b) {
+      return a / b;
+    } else {
+      return b / a;
+    }
+  }
+
+  const smallestResourceThatsLarger = function( annotatedCandidates ) {
+    const distances = annotatedCandidates.map( c => c.linearDistance );
+    const larger = distances.filter( (d) => { return d >= 0 } );
+    let selectedDistance;
+    if ( larger.length === 0 ) {
+      selectedDistance = Math.max( ...distances );
+    } else {
+      selectedDistance = Math.min( ...larger );
+    }
+    return annotatedCandidates.find( c => c.linearDistance === selectedDistance );
+  }
+
+  const closestResource = function( annotatedCandidates ) {
+    const distances = annotatedCandidates.map( c => c.geometricDistance );
+    const minDistance = Math.min( ...distances );
+    return annotatedCandidates.find( c => c.geometricDistance === minDistance );
+  }
+
+  const selectResourceFromSrcset = function( annotatedCandidates, dpr ) {
+    if ( dpr > 1 ) {
+      return closestResource( annotatedCandidates );
+    } else {
+      return smallestResourceThatsLarger( annotatedCandidates );
+    }
+  }
+
+  // main
+  // only run if there is an (explicit or implied) sizes value, and more than one resource in the srcset
+  // (ie, if sizes errors could matter)
+  if ( imgData.sizesWidth && srcsetCandidates && srcsetCandidates.length > 1 ) {
+
+    // determine the effective densities of the srcset resources
+    // using ideal (rather than actual) sizes value
+
+    // deep copy
+    idealSizesSrcsetCandidates = JSON.parse( JSON.stringify( srcsetCandidates ) );
+    // modify in place (just like before) TODO: turn this into a function, use in both places...
+    // overwriting densities that we copied
+    idealSizesSrcsetCandidates.forEach( i => {
+      if ( i.hasOwnProperty( 'd' ) ) {
+        i.density = i.d;
+      } else if ( i.hasOwnProperty( 'w' ) && imgData.clientWidth > 0 ) {
+        i.density = i.w / imgData.clientWidth;
+      } else {
+        i.density = 1;
+      }
+      // in addition to densities, annotate candidates with distances from DPR, used in srcset selection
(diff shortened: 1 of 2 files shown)


rviscomi:
I might not have time for a deep code review, but in the meantime @eeeps could you look at some of the linter errors?

eeeps:
While running this through a suite of actual pages, I found an issue where if `w` descriptors in the `srcset` were wrong (e.g.—on netflix.com's hero right now!), this metric attributed that error to the `sizes` attribute.

Addressing that now...

eeeps:
Ok, here are some WPT results, showing behavior in different contexts:

- [css-tricks.com](https://www.webpagetest.org/result/220527_AiDcEY_H12/1/details/): small sizes errors that do affect srcset selection
- [harvard.edu](https://www.webpagetest.org/result/220527_BiDcPS_G23/1/details/): large sizes errors that do not affect srcset selection (because we were already picking the largest resource)
- [netflix.com](https://www.webpagetest.org/result/220527_BiDc8G_FJ5/2/details/): Erroneous w descriptors on hero image that are accounted for separately
- And a few pages that show no disruption to existing reporting, in contexts where these metrics don't matter:
  - [nytimes.com](https://www.webpagetest.org/result/220527_BiDcWM_G2G/1/details/) (no w descriptors)
  - [shutterstock.com](https://www.webpagetest.org/result/220527_BiDcR9_G35/1/details/) (no w descriptors)
  - [mailchimp.com](https://www.webpagetest.org/result/220527_AiDcEA_H0Z/1/details/) (shows interaction with lazy loading — we don’t get metrics, even though there are sizes, as expected)

dist/responsive_images.js
```diff
@@ -478,6 +482,13 @@ function getImgData( img ) {
     imgData.bitsPerPixel = ( imgData.byteSize * 8 ) / ( imgData.approximateResourceWidth * imgData.approximateResourceHeight );
   }
 
+  // if there was a w descriptor -- was it accurate?
+  if ( imgData.currentSrcWDescriptor &&
+      imgData.approximateResourceWidth ) {
+    imgData.wDesciptorAbsoluteError = imgData.currentSrcWDescriptor - imgData.approximateResourceWidth;
+    imgData.wDesciptorRelativeError = imgData.wDesciptorAbsoluteError / imgData.approximateResourceWidth;
```
rviscomi, line 489:
```suggestion
    imgData.wDescriptorAbsoluteError = imgData.currentSrcWDescriptor - imgData.approximateResourceWidth;
    imgData.wDescriptorRelativeError = imgData.wDescriptorAbsoluteError / imgData.approximateResourceWidth;
```

dist/responsive_images.js
```diff
@@ -499,6 +510,90 @@ function getImgData( img ) {
     Object.values( imgData.intrinsicOrExtrinsicSizing ).includes( 'intrinsic' )
   );
 
+  // what was the impact of incorrect sizes values?
+
+  // helpers...
+  function geometricDistance(a, b) {
```
yoavweiss, line 516: In Chromium I used a [geometric mean](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/html/parser/html_srcset_parser.cc;l=396?q=geometric%20srcset%20f:.cc&ss=chromium) *between the candidates* as a tie breaker. I think this is slightly different

dist/responsive_images.js
```diff
@@ -499,6 +510,90 @@ function getImgData( img ) {
...
+    if ( larger.length === 0 ) {
+      selectedDistance = Math.max( ...distances );
+    } else {
+      selectedDistance = Math.min( ...larger );
+    }
+    return annotatedCandidates.find( c => c.linearDistance === selectedDistance );
+  }
+
+  const closestResource = function( annotatedCandidates ) {
+    const distances = annotatedCandidates.map( c => c.geometricDistance );
```
yoavweiss, line 537: `return annotatedCandidates.sort((a, b) => {return a.geometricDistance - b.geometricDistance;})[0]` assuming annotatedCandidates is not empty (we should probably validate that assumption)

dist/responsive_images.js
yoavweiss: Specific example where this would give you the wrong results:
1x and 2x candidates have a geometric mean of 1.44, so a 1.42 DPR screen would pick the 1x candidate, while this method would give you a ratio of 1.42 for the 1x and 1.408 for the 2x. I'm not sure if there are examples where the diff is more likely to be significant, but I wonder why you deviated from the geometric mean. Any particular reason?

dist/responsive_images.js
```diff
@@ -499,6 +510,90 @@ function getImgData( img ) {
...
+  // helpers...
+  function geometricDistance(a, b) {
+    if (a > b) {
+      return a / b;
+    } else {
+      return b / a;
+    }
+  }
+
+  const smallestResourceThatsLarger = function( annotatedCandidates ) {
```
yoavweiss, line 524: `smallestResourceThatsLargerThanDPR`?

dist/responsive_images.js
```diff
@@ -499,6 +510,90 @@ function getImgData( img ) {
...
+    const distances = annotatedCandidates.map( c => c.linearDistance );
```
yoavweiss, line 525:
```
const larger = annotatedCandidates.filter(c => return c.linearDistance >=0);
if (larger.size === 0) {
  if (annotatedCandidates.length === 0) {
    // bail
  }
  return annotatedCandidates.sort((a, b) => {return a.linearDistance - b.linearDistance;})[annoratedCandidates.length - 1];
} 
return larger.sort((a, b) => {return a.linearDistance - b.linearDistance;})[0];
```

rviscomi:
@tunetheweb I'm not sure if an HA admin needs to be around to merge PRs but if so could you keep an eye on this while I'm away this week?

tunetheweb:
Aye. Though travelling myself this week. But much shorter plane flights than you!

@eeeps if you could get the feedback addressed today that would be great!

dist/responsive_images.js
eeeps: I twisted myself in a couple of knots until arriving on this solution, which was pleasingly functional and didn't require re-sorting, or special-casing for the first or last element. Glad it actually works! But I probably should have looked at the Chromium source and just copied that, first...

dist/responsive_images.js
eeeps: Good point: "larger than what?" I'll rename.

dist/responsive_images.js
eeeps: Empty annotatedCandidates is proobbbably not possible because of https://github.com/eeeps/custom-metrics/blob/016611994161bc19cca6d2b5020ae3b19eb13d3e/dist/responsive_images.js#L553, but I'll add checks. And yeah your logic is cleaner / more standard I think...

eeeps:
@tunetheweb @yoavweiss I think I addressed all of Yoav's feedback but I would like to re-run a few validation tests to make sure that I didn't break anything in the process. I need to run at the moment, I'll try to get back to it in a few hours and give @tunetheweb a final thumbs up.

Thank you!!

dist/responsive_images.js
```diff
@@ -499,6 +510,97 @@ function getImgData( img ) {
...
+        return a.linearDistance - b.linearDistance;
+      } )[ annotatedCandidates.length - 1 ];
+    }
+    return larger.sort( ( a, b ) => {
+      return a.linearDistance - b.linearDistance;
+    } )[ 0 ];
+  }
+
+  const resourceThatsClosestToDPR = function( annotatedCandidates ) {
+    if ( !annotatedCandidates || !annotatedCandidates.length || annotatedCandidates.length === 0 ) {
```
yoavweiss, line 540: Is the zero equivalence check really needed after checking `!length`?

dist/responsive_images.js
```diff
@@ -499,6 +510,97 @@ function getImgData( img ) {
...
+  function geometricDistance(a, b) {
+    if (a > b) {
+      return a / b;
+    } else {
+      return b / a;
+    }
+  }
+
+  const smallestResourceThatsLargerThanDPR = function( annotatedCandidates ) {
+    if ( !annotatedCandidates || !annotatedCandidates.length || annotatedCandidates.length === 0 ) {
```
yoavweiss, line 525: Same question as below

yoavweiss (approved):
LGTM

dist/responsive_images.js
eeeps: Nope! I'll remove.

dist/responsive_images.js
eeeps: Same answer.

eeeps:
Some updated test results, with @rviscomi and @yoavweiss 's changes:

- [Harvard](https://www.webpagetest.org/result/220531_BiDc9Q_42P/1/details/)
- [CSS-Tricks](https://www.webpagetest.org/result/220531_AiDc6A_3ZZ/1/details/)
- [Netflix](https://www.webpagetest.org/result/220531_AiDcY1_405/1/details/)
- [Shutterstock](https://www.webpagetest.org/result/220531_BiDc86_43N/1/details/)
- [Mailchimp](https://www.webpagetest.org/result/220531_BiDcSG_43T/1/details/)

Diffing vs the earlier results, the only changes I see are:

1. changed content on the pages
2. "wDescripor" → "wDescriptor"

Which is as I'd expect. So.... I think we're good to merge, @tunetheweb

tunetheweb (approved):
Thanks @yoavweiss and @eeeps 

Ran it on a couple more tests and LGTM!

PR #33, opened 2022-05-27
Inject worker

kevinfarrugia:
Required for [JavaScript 2022](https://github.com/HTTPArchive/almanac.httparchive.org/issues/2880)

Added support to count the number of times the Worker constructor is called. Code is setup to support additional constructors if needed.

Test: https://webpagetest.httparchive.org/result/220527_E9_5/1/details/#waterfall_view_step1

Uses custom metric:
```
[worker]
return JSON.stringify(httparchive_observers.constructor_stacks["Worker"]);
```

Results include:

worker | 1
-- | --


kevinfarrugia:
Replaced with #35

PR #35, opened 2022-05-27
Observe Worker initialization

kevinfarrugia:
Required for [JavaScript 2022](https://github.com/HTTPArchive/almanac.httparchive.org/issues/2880)

Added support to count the number of times the Worker constructor is called. 

Test: https://webpagetest.httparchive.org/result/220529_6B_M/1/details/#waterfall_view_step1

Uses custom metric:
```
[observers]
return Object.fromEntries(Object.entries(httparchive_observers).filter(([_, value]) => value));
```

Results include:

observers | { Worker: 1 }
-- | --

### DIFF
--- a/dist/javascript.js
+++ b/dist/javascript.js
@@ -94,6 +94,6 @@ return JSON.stringify({
     })(),
 
     workers: (() => {
-        return httparchive_observers.constructor_stacks["Worker"];
+        return httparchive_observers["Worker"];
     })()
 });
--- a/dist/javascript.js
+++ b/dist/javascript.js
@@ -92,4 +92,5 @@ return JSON.stringify({
 
         return noscript_data;
     })()
+
 });
--- a/dist/javascript.js
+++ b/dist/javascript.js
@@ -91,6 +91,9 @@ return JSON.stringify({
         };
 
         return noscript_data;
-    })()
+    })(),
 
+    workers: (() => {
+        return httparchive_observers.constructor_stacks["Worker"];
+    })()
 });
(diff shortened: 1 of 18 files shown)


rviscomi:
I think all that's needed is a minimal change to add `Worker` to the `OBSERVERS` list. There was a bug in the implementation that I recently fixed in #32 where top-level properties under `window` weren't being observed properly. That might by why you thought you needed a separate implementation for constructors.

Here's a [test](https://webpagetest.httparchive.org/result/220528_MP_4/1/details/#waterfall_view_step1) where I just added the constructor name to the const, and it worked.

image

kevinfarrugia:
Thank you @rviscomi . I have updated the PR and it seems to work correctly.

rviscomi:
Did you push your changes? I'm still seeing the separate `CONSTRUCTORS` array.

kevinfarrugia:
😴. Updated.

dist/javascript.js
```diff
@@ -91,6 +91,9 @@ return JSON.stringify({
         };
 
         return noscript_data;
-    })()
+    })(),
 
+    workers: (() => {
```
rviscomi, line 96: Similarly I think we can rely on the `observers.js` custom metric to extract this value. Otherwise LGTM.

rviscomi (changes requested):

kevinfarrugia:
Updated test using `observers.js`.

https://webpagetest.httparchive.org/result/220529_6B_M/1/details/#waterfall_view_step1

rviscomi (approved):

PR #39, opened 2022-05-29
Add Soft Metrics, Revisions

mel-ada:
**Soft Metrics:**
- Add `Strict` suffix to existing metrics
- Add `Soft` suffixed version of metrics with more relaxed requirements
- [Not Working] Add function `doesElementCoverPercentageOfViewport`. At the moment, this uses IntersectionObserver, which is working in the browser. I think IntersectionObserver is not supported in `$WPT_BODIES` because it is, in fact, not a browser. The alternative is using `getBoundingClientRect` and the viewport size to compute % of viewport, however I haven't had time to research/don't know of a way to get viewport size that is supported in `$WPT_BODIES`.
- [Not Tested] Add function `findHighestZ` to find the highest z-index in the document. This works in the browser. I think that since the prior function is not working, this one isn't running either.

**Revisions:**
- Default isLcpStaticallyDiscoverable to `true`. I noticed that this is only ever evaluated if the LCP has an external resource, which means that we would be counting sites with LCP elements that are divs/etc as not statically discoverable, even though they don't have a resource in the first place.
- Removed scan for `svg` LCP element hack, because the hack referred to in [the article](https://www.devisedlabs.com/blog/largest-contentful-paint-lcp-hack) is actually an `img` element with the `src` attribute set to an svg.

### DIFF
--- a/dist/performance.js
+++ b/dist/performance.js
@@ -24,8 +24,10 @@ function getLcpElement() {
             resolve(naiveLcpEntry);
         }).observe({ type: "largest-contentful-paint", buffered: true });
     }).then(({ startTime, element, url, size, loadTime, renderTime }) => {
+        const cover90viewport = false;
+        doesElementCoverPercentageOfViewport(element, 90);
         const attributes = getAttributes(element);
-        const styles = getComputedStyles(element, ['background-image']);
+        const styles = getComputedStyles(element, ['background-image', 'pointer-events', 'position', 'width', 'height']);
         return {
             startTime,
             nodeName: element?.nodeName,
@@ -37,7 +39,8 @@ function getLcpElement() {
             boundingClientRect: element?.getBoundingClientRect().toJSON(),
             naturalWidth: element?.naturalWidth,
             naturalHeight: element?.naturalHeight,
-            styles
+            styles,
+            cover90viewport
         };
     });
 }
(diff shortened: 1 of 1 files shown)


dist/performance.js
```diff
@@ -171,24 +160,70 @@ function getGamingMetrics(rawDoc) {
...
+// Source: https://stackoverflow.com/questions/57786082/determine-how-much-of-the-viewport-is-covered-by-element-intersectionobserver
+// percentage is a whole number (ex: 90, not .9)
+const doesElementCoverPercentageOfViewport = (element, percentage) => {
+
+    const observer = new IntersectionObserver(function (entries) {
+        entries.forEach(function (entry) {
+            const entryBCR = entry.target.getBoundingClientRect();
+            const percentOfViewport = ((entryBCR.width * entryBCR.height) * entry.intersectionRatio) / ((window.innerWidth * window.innerHeight) / 100);
+            if (percentOfViewport > percentage) {
+                cover90viewport = true;
```
mel-ada, line 185: Yes this variable is in the outer scope and has 2 different instances, ouch.

dist/performance.js
```diff
@@ -171,24 +160,70 @@ function getGamingMetrics(rawDoc) {
...
+            }
+        });
+    },);
```
rviscomi, line 188:
```suggestion
    });
```

dist/performance.js
rviscomi: Does this write to global scope? I wouldn't expect the other instance to be writable from this scope.

Since we know the scroll position will always be 0, is it possible to avoid the need for IntersectionObserver by looking at `entryBCR`'s x and y values, then immediately returning the boolean?

dist/performance.js
```diff
@@ -171,24 +160,70 @@ function getGamingMetrics(rawDoc) {
...
                     styleObj['left'] == '0px' &&
                     styleObj['z-index'] == '999'){
-                returnObj['fidIframeOverlay'] = true;
+                returnObj['fidIframeOverlayStrict'] = true;
             }
         }
+
+        let cover90viewport = false;
+        doesElementCoverPercentageOfViewport(iframeElement, 90)
+        returnObj['fidIframeOverlaySoft'] = cover90viewport && styleObj['z-index'] === findHighestZ(rawDoc);
```
rviscomi, line 169: Since this is the "soft" version, WDYT about any positive z-index to be indicative that the intent was to overlay the iframe? Your way is more like the strict interpretation even if not the literal z-index used by the blog post.

25prathamesh: @mel-ada I think we should not use the z-index check in the soft version.   someone can achieve an iframe overlay without a z-index by properly positioning the iframe & elements.  I have added it earlier because I found some iframe overlay examples were using that property

mel-ada:
@rviscomi @25prathamesh @tunetheweb I've addressed the feedback and added a link to a working WPT, this is ready for another look!

tunetheweb:
Ah thanks. Shall we merge this then? Can always pick up changes on the other branch.

25prathamesh:
HI @tunetheweb  
I checked the current code we are adding a key in gaming_metrics JSON only if that hack is detected on page except lcpOverlaySoft , for lcpOverlaySoft its true/false need to be fixed.

I am running all gaming metrics tests some are failed that's need to be checked, You can merge current changes we will push fixes on other branch

WPT tests results  
Detect chrome/pagespeed/gtmetric  user agent hack  (code inside external JS) - DETECTION PASSED
https://www.webpagetest.org/result/220531_AiDcSX_91R/1/details/#waterfall_view_step1

Detect chrome user agent hack (inline JS)  - DETECTION FAILED 
https://www.webpagetest.org/result/220531_BiDcJ6_8ZW/1/details/#waterfall_view_step1

Fid Iframe Overlay -  DETECTION PASSED
https://www.webpagetest.org/result/220531_BiDc39_8YF/1/details/#waterfall_view_step1

PR #40, opened 2022-05-29
Added custom metric for Webpack and ParcelJS

kevinfarrugia:
Added custom metric to return array of known bundlers in use on a page.

**Webpack**
- https://webpagetest.httparchive.org/result/220529_JG_E/1/details/#waterfall_view_step1
- https://webpagetest.httparchive.org/result/220529_RD_C/1/details/#waterfall_view_step1

javascript_new|{"bundler":["webpack"]}
--|--

**Parcel**
- https://webpagetest.httparchive.org/result/220529_SW_J/1/details/#waterfall_view_step1

javascript_new|{"bundler":["parcel"]}
--|--

**None**
- https://webpagetest.httparchive.org/result/220529_FY_F/1/details/#waterfall_view_step1

javascript_new|{"bundler":[]}
--|--

Note that this implementation is built on [wappalyzer](https://github.com/wappalyzer/wappalyzer)'s implementation and is only able to recognize Webpack and Parcel.

### DIFF
--- a/dist/javascript.js
+++ b/dist/javascript.js
@@ -91,6 +91,14 @@ return JSON.stringify({
         };
 
         return noscript_data;
-    })()
+    })(),
 
+    bundler: (() => {
+        const bundler = {
+            webpack: !(typeof webpackJsonp === "undefined" && typeof webpackChunk === "undefined"),
+            parcel: !(typeof parcelRequire === "undefined")
+        }
+
+        return Object.entries(bundler).filter(n => n[1]).map(n => n[0]);
+    })()
 });


dist/javascript.js
```diff
@@ -91,6 +91,14 @@ return JSON.stringify({
         };
 
         return noscript_data;
-    })()
+    })(),
 
+    bundler: (() => {
+        const bundler = {
+            webpack: !(typeof webpackJsonp === "undefined" && typeof webpackChunk === "undefined"),
+            parcel: !(typeof parcelRequire === "undefined")
```
rviscomi, line 99: Just to confirm, when detected these are all expected to be in the global scope?

Also I only see `webpackJsonp` in the Wappalyzer implementation.

kevinfarrugia: Either or. So for Webpack, either `webpackJsonp` or `webpackChunk`. I created a [PR](https://github.com/wappalyzer/wappalyzer/pull/6309) for Wappalyzer but it is still awaiting review. 

It is also not a 100% solution because I have come across implementations where the global variable has a different name, ex: `webpackChunkMyProject`.

rviscomi: Thanks for clarifying. I suppose it's possible to get all properties on the window object and test if any are prefixed with `webpackChunk` but that may be overkill. This is ok to merge.

rviscomi (approved):

PR #41, opened 2022-05-29
Fetch & parse sourceMaps to check for @babel transpiler

kevinfarrugia:
Checking for `@babel` by parsing and downloading the source maps. Limited to pages which have publicly available source maps.

Test 1 (with public sourcemaps): https://webpagetest.httparchive.org/result/220529_QN_1J/1/details/#waterfall_view_step1

```
  "sourceMaps": {
    "count": 4,
    "ext": {
      "scss": 123,
      "js": 541,
      "svg": 13,
      "jpg": 1,
      "png": 7,
      "jsx": 152
    },
    "babel": true,
    "js": { "size": 1765807, "stats": {} }
  }
```

Test 2 (without public sourcemaps): https://webpagetest.httparchive.org/result/220529_37_1K/

```
  "sourceMaps": {
    "count": 17,
    "ext": {},
    "js": { "size": 0, "stats": {} }
  }
```

Test 3 (without sourcemaps): https://webpagetest.httparchive.org/result/220529_FV_1M/

```
{ "sourceMaps": { "count": 0, "ext": {} } }
```

### DIFF
--- a/dist/javascript.js
+++ b/dist/javascript.js
@@ -94,7 +94,6 @@ let getSourceMaps = (async () => {
 
         return url;
       });
-        
       if(babel > 0) {
         results.babel = true;
       }
--- a/dist/javascript.js
+++ b/dist/javascript.js
@@ -1,4 +1,4 @@
-// [javascript]
+//[javascript]
 // Uncomment the previous line for testing on webpagetest.org
 
 // Please, refer to instructions for adding a custom metric in almanac.js.
--- a/dist/javascript.js
+++ b/dist/javascript.js
@@ -272,4 +272,5 @@ return Promise.all([getSourceMaps]).then(([sourceMaps]) =>
 
         return Object.entries(bundler).filter(n => n[1]).map(n => n[0]);
     })()
-});
+  })
+);
--- a/dist/javascript.js
+++ b/dist/javascript.js
@@ -1,4 +1,4 @@
-[javascript]
+// [javascript]
 // Uncomment the previous line for testing on webpagetest.org
 
 // Please, refer to instructions for adding a custom metric in almanac.js.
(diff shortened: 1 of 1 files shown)


dist/javascript.js
```diff
@@ -3,94 +3,266 @@
...
+          }
+        }
+
+        if (url.pathname.match(/@babel/)) {
+            babel++;
+        }
+
+        return url;
+      });
+
```
rviscomi, line 97: fix linter
```suggestion

```

dist/javascript.js
```diff
@@ -3,94 +3,266 @@
...
+    setTimeout(() => {controller.abort()}, 5000);
+    return fetch(url, {signal: controller.signal});
+  }
+
 const requests = $WPT_REQUESTS;
 
-return JSON.stringify({
+let getSourceMaps = (async () => {
+  const $$ = (s) => [...document.querySelectorAll(s)];
+  const sourcemapRegex = /\/\/[#@] sourceMappingURL=(.+?)$/;
```
rviscomi, line 16: Do you also want to capture the [`SourceMap`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/SourceMap) header?

kevinfarrugia: Is there a custom-metric already looking at the SourceMap header? From last year's chapter, usage is low with 0.1% of pages using this.

https://almanac.httparchive.org/en/2021/javascript#source-maps

rviscomi: Here's an example:

https://github.com/HTTPArchive/custom-metrics/blob/main/dist/responsive_images.js#L458

I agree usage is probably too low to matter, so I'll leave it up to you.

rviscomi (approved):
Will merge as-is to ensure these changes make it into the crawl. Open a new PR if needed.

dist/javascript.js
kevinfarrugia: Yeah, will leave out for now.

PR #46, opened 2022-06-23
Fix LCP metrics

### LINKED ISSUE #45: Issues with some new performance custom metrics
I was taking a look at some of the custom metrics added in #20 and I noticed a few issues with the code:

1. [Line 260](https://github.com/HTTPArchive/custom-metrics/blob/48865343b15d898b3c472d72a1327e50333f24cc/dist/performance.js#L260) creates a URL from a `srcset` attribute, but `srcset` attributes contain multiple URLs that need to be parsed and iterated over, so this will never match. (See [this false negative](https://webpagetest.httparchive.org/result/220609_DxI7_3WGSH/) for an example failure case).
2. [Line 259](https://github.com/HTTPArchive/custom-metrics/blob/48865343b15d898b3c472d72a1327e50333f24cc/dist/performance.js#L259) only looks at `srcset` attributes in `<source>` elements, but images can also have `srcset` attributes.
3. [Line 265](https://github.com/HTTPArchive/custom-metrics/blob/48865343b15d898b3c472d72a1327e50333f24cc/dist/performance.js#L265): Images in preload tags are discoverable from the main document ([preload scanner](https://web.dev/preload-scanner/)), so I think this condition needs to be inclusive of cases where `isLcpPreloaded` is true.
4. [Line 267](https://github.com/HTTPArchive/custom-metrics/blob/48865343b15d898b3c472d72a1327e50333f24cc/dist/performance.js#L267): `<preload>` tags can also contain an `imagesrcset` attribute, so that needs to be checked as well (and it needs to be parsed, like in (1) above).
5. Also [Line 267](https://github.com/HTTPArchive/custom-metrics/blob/48865343b15d898b3c472d72a1327e50333f24cc/dist/performance.js#L267): preload hints can be set on HTTP headers, so just checking `<link>` element is not enough to know whether or not a resource was preloaded.

In addition to these issues, I'm also concern because (if I understand correctly) the custom metric code runs at the end of a test on the live DOM, which may not catch many of the antipatterns we're seeing around things like JS lazy loading of the LCP image.

For example, if an inline JS script replaces `data-src` attributes with `src` attributes for images determined to be within the viewport, I think this custom metric code would falsely determine that that resource was discoverable, which it actually wasn't, correct?

rviscomi:
Fixes #45 

- [x] Iterate over multiple `srcset` values
- [x] Handle `img[srcset]` cases
- [x] Include preload in discoverability check
- [x] Handle `srcset` in preload cases
- [x] Include HTTP preloads in preload check

### DIFF
--- a/dist/performance.js
+++ b/dist/performance.js
@@ -227,6 +227,15 @@ function calcOcclusion(rect){
     return Math.max(0, calcArea(clipped_rect)/calcArea(rect))
 }
 
+function splitSrcSet(srcset, baseUrl) {
+    // "img.jpg 100w, img2.jpg 300w"
+    return srcset.split(',').map(srcDesc => {
+        // "img.jpg 100w", " img2.jpg 300w"
+        const src = srcDesc.trim().split(' ')[0];
+        return new URL(src, baseUrl).href;
+    });
+}
+
 return Promise.all([getLcpElement()]).then(([lcp_elem_stats]) => {
     const lcpUrl = lcp_elem_stats.url;
     const rawDoc = getRawHtmlDocument();
@@ -256,8 +265,8 @@ return Promise.all([getLcpElement()]).then(([lcp_elem_stats]) => {
         // Check if LCP resource reference is in the raw HTML (as opposed to being injected later by JS)
         rawLcpElement = Array.from(rawDoc.querySelectorAll('picture source, img')).find(i => {
             let src = i.src;
-            if (i.nodeName == 'SOURCE') {
-                src = new URL(i.srcset, location.href).href;
+            if (i.hasAttribute('srcset')) {
+                src = splitSrcSet(i.srcset, location.href).find(src => src == lcpUrl);
             }
 
             return src == lcpUrl;
--- a/dist/performance.js
+++ b/dist/performance.js
@@ -227,13 +227,32 @@ function calcOcclusion(rect){
     return Math.max(0, calcArea(clipped_rect)/calcArea(rect))
 }
 
-function splitSrcSet(srcset, baseUrl) {
+function splitSrcSet(srcset) {
     // "img.jpg 100w, img2.jpg 300w"
     return srcset.split(',').map(srcDesc => {
         // "img.jpg 100w", " img2.jpg 300w"
         const src = srcDesc.trim().split(' ')[0];
-        return new URL(src, baseUrl).href;
+        return new URL(src, location.href).href;
     });
+}
+
+function parseLinkHeader(link) {
+    const srcPattern = /<([^>]+)>/;
+    const paramPattern = /([^=]+)=['"]?([^'"]+)['"]?/;
+    return Object.fromEntries(link.split('\n').map(l => {
+        let [src, ...params] = l.split(';');
+
+        if (!srcPattern.test(src)) {
+            return [];
+        }
+        src = src.match(srcPattern)[1];
+
+        params = params.map(p => {
+            const [_, key, value] = p.trim().toLowerCase().match(paramPattern);
+            return {key, value};
+        });
+        return [src, params];
+    }));
 }
 
 return Promise.all([getLcpElement()]).then(([lcp_elem_stats]) => {
@@ -266,24 +285,33 @@ return Promise.all([getLcpElement()]).then(([lcp_elem_stats]) => {
         rawLcpElement = Array.from(rawDoc.querySelectorAll('picture source, img')).find(i => {
             let src = i.src;
             if (i.hasAttribute('srcset')) {
-                src = splitSrcSet(i.srcset, location.href).find(src => src == lcpUrl);
+                src = splitSrcSet(i.srcset).find(src => src == lcpUrl);
             }
 
             return src == lcpUrl;
         });
-        isLcpPreloaded = !!Array.from(rawDoc.querySelectorAll('link')).find(link => {
+        isLcpPreloaded = Array.from(rawDoc.querySelectorAll('link')).some(link => {
             if (link.rel != 'preload') {
                 return false;
             }
 
             let src = link.href;
             if (link.hasAttribute('imagesrcset')) {
-                src = splitSrcSet(link.imagesrcset, location.href).find(src => src == lcpUrl);
+                src = splitSrcSet(link.imagesrcset).find(src => src == lcpUrl);
             }
 
             return src == lcpUrl;
         });
-        isLcpStaticallyDiscoverable = !!rawLcpElement || !!isLcpPreloaded;
+        let isLcpPreloadedInHeaders = false;
+        const linkHeader = response_bodies[0].response_headers.link;
+        if (linkHeader) {
+            const directives = parseLinkHeader(linkHeader);
+            isLcpPreloadedInHeaders = lcpUrl in directives && directives[lcpUrl].some(param => {
+                return param.key == 'rel' && param.value == 'preload';
+            });
+        }
+        isLcpPreloaded = isLcpPreloaded || isLcpPreloadedInHeaders;
+        isLcpStaticallyDiscoverable = !!rawLcpElement || isLcpPreloaded;
         responseObject = response_bodies.find(r => {
             return r.url == lcpUrl;
         });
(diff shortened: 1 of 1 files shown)


rviscomi:
alistapart: https://webpagetest.httparchive.org/result/220623_N9_2/1/details/#waterfall_view_step1

rviscomi:
Added an `lcp_preload` object that will include `Link` header preload directives and `<link>` tag preload attributes when available. See tests in previous comment.

dist/performance.js
```diff
@@ -227,60 +239,141 @@ function calcOcclusion(rect){
...
+        }
+
+        return src == lcpUrl;
+    });
+
+    if (preloadTag) {
+        lcpPreload.tag = summarizeElement(preloadTag);
+    }
+
+    const linkHeaderString = response_bodies[0].response_headers.link;
```
philipwalton, line 299: I'm not sure what the format of the `response_headers` object is, but it's definitely possible for a page to contain more than 1 `Link` header, so it's probably worth double-checking that this data structure supports that case.

dist/performance.js
```diff
@@ -227,60 +239,141 @@ function calcOcclusion(rect){
...
-        });
-        responseObject = response_bodies.find(r => {
-            return r.url == lcpUrl;
-        });
-        if (responseObject) {
-            // Don't write the response body to custom metrics.
-            responseObject.response_body = undefined;
-        }
+        isLcpPreloaded = 'header_directives' in lcpPreload || 'tag' in lcpPreload;
+        isLcpStaticallyDiscoverable = !!rawLcpElement || isLcpPreloaded;
```
philipwalton, line 367: Optional nit: I'm not sure the `!!` is needed here, and I think it makes it a bit less readable.

philipwalton (changes requested):

rviscomi:
@philipwalton @tunetheweb does this look ready to merge?

dist/performance.js
```diff
@@ -227,60 +239,141 @@ function calcOcclusion(rect){
...
+        lcpPreload.header_directives = Object.fromEntries(directives[lcpUrl].map(d => {
+            return [d.key, d.value];
+        }));
+    }
+
+    return lcpPreload;
+}
+
+function getRawLcpElement(rawDoc, lcpUrl) {
+    const rawLcpElement = Array.from(rawDoc.querySelectorAll('picture source, img')).find(i => {
```
tunetheweb, line 323: This doesn't get the rawLCPElement, it gets the rawImageLCPElement.
Should it return the text node for text LCP elements? If not should it early return if lcpURL is null/empty/undefined?

Also should it include `<image>` elements from SVGs?

Also not sure how it should handle "An element with a background image loaded via the [url()](https://developer.mozilla.org/docs/Web/CSS/url()) function (as opposed to a [CSS gradient](https://developer.mozilla.org/docs/Web/CSS/CSS_Images/Using_CSS_gradients))"

dist/performance.js
philipwalton: This comment was resolved, but can you confirm that this is supported?

I'm thinking about a case where a page responds with multiple `Link: ...` headers:

```
Link: <script1.js>;rel="preload";as="script"
Link: <script2.js>;rel="preload";as="script"
```

Does the `response_headers` object merge those into a single value?

dist/performance.js
```diff
@@ -227,60 +239,141 @@ function calcOcclusion(rect){
...
+        let src = i.src;
+        if (i.hasAttribute('srcset')) {
+            src = splitSrcSet(i.srcset).find(src => src == lcpUrl);
+        }
+
+        return src == lcpUrl;
```
tunetheweb, line 329: If you are not adding text elements, then this might be null/none/undefined.

```suggestion
        return src && src == lcpUrl;
```

dist/performance.js
```diff
@@ -227,60 +239,141 @@ function calcOcclusion(rect){
...
-        });
-        responseObject = response_bodies.find(r => {
-            return r.url == lcpUrl;
-        });
-        if (responseObject) {
-            // Don't write the response body to custom metrics.
-            responseObject.response_body = undefined;
-        }
+        isLcpPreloaded = 'header_directives' in lcpPreload || 'tag' in lcpPreload;
+        isLcpStaticallyDiscoverable = !!rawLcpElement || isLcpPreloaded;
```
tunetheweb, line 367: Again this should probably be better named, or return true if it's a text node.

tunetheweb (changes requested):
Doesn't look like it works too well for text nodes.

For example: https://www.webpagetest.org/result/220628_BiDcJC_EAW/2/details/#waterfall_view_step1

```
    "raw_lcp_element": null,
    "lcp_resource": null,
    "is_lcp_statically_discoverable": true,
    "is_lcp_preloaded": null,
    "lcp_preload": null,
```

- `raw_lcp_element` is `null` but perhaps should be set to the text node?
- `is_lcp_statically_discoverable` is true but more due to a bug in how rawLcpElement is calced, than by intention and won't be for all text.
- `is_lcp_preloaded` is `null` but probably should be false in this case.

dist/performance.js
tunetheweb: Yes it does.

For example alistpart.com has the following:

image

And using this, simplified, custom-metric:

```js
[link-headers]

function parseLinkHeader(link) {
    if (!link) {
        return {};
    }

    const srcPattern = /<([^>]+)>/;
    const paramPattern = /([^=]+)=['"]?([^'"]+)['"]?/;
    return Object.fromEntries(link.split(/[,\n]/).map(l => {
        let [src, ...params] = l.split(';');

        if (!srcPattern.test(src)) {
            return [];
        }
        src = src.match(srcPattern)[1];
        src = new URL(src, location.href).href;

        params = params.map(p => {
            p = p.trim().toLowerCase();
            if (!paramPattern.test(p)) {
                return;
            }
            const [_, key, value] = p.match(paramPattern);
            return {key, value};
        });
        return [src, params];
    }));
}

const response_bodies = $WPT_BODIES;
return parseLinkHeader(response_bodies[0].response_headers.link);
```

We see the following [in this example test](https://www.webpagetest.org/result/220628_AiDcRT_ERX/2/details/#waterfall_view_step1):

```json
{
    "https:\/\/alistapart.com\/wp-json\/": [
        {
            "key": "rel",
            "value": "https:\/\/api.w.org\/"
        }
    ],
    "https:\/\/alistapart.com\/wp-json\/wp\/v2\/pages\/6000475": [
        {
            "key": "rel",
            "value": "alternate"
        },
        {
            "key": "type",
            "value": "application\/json"
        }
    ],
    "https:\/\/wp.me\/PaHlEW-paZR": [
        {
            "key": "rel",
            "value": "shortlink"
        }
    ]
}
```

So you can see it picked up all three link headers and `parseLinkHeader` correctly separated them out.

rviscomi:
@tunetheweb given that the current analysis is only concerned with image-based LCP elements, do you consider your feedback merge-blocking? Are there any minor changes like todo comments or non-breaking variable renames that would satisfy your concerns?

dist/performance.js
philipwalton: Great, thanks for confirming!

dist/performance.js
rviscomi: This function is only called when `lcpUrl != ''` (see `isLcpExternalResource`)

tunetheweb:
I'm OK with either adding TODOs or renaming to make sure more obvious it's just media. Personally I'd go with the TODOs as think it would be good to get text in here at some point so we could say "X % of LCP elements are in initial HTML" if we wanted to in future, rather than just "X % of external LCP sources are in initial HTML". Maybe even could update `is_lcp_statically_discoverable` to include where `lcpUrl` is none to include that in this PR?

I do think the line 329 suggestion should be done though as I discovered a false positive with it, in the example I included.

dist/performance.js
rviscomi: Not sure what you're suggesting return true. Could you clarify? If LCP is text `isLcpStaticallyDiscoverable` defaults to true, and `isLcpExternalResource` is false so this code isn't executed.

rviscomi (commented):
Sorry @philipwalton I just realized I had pending comments to your feedback that never got sent. Including them now in my replies to @tunetheweb's feedback.

dist/performance.js
tunetheweb: Ah missed that. Resolving.

tunetheweb (approved):
LGTM. Sorry for earlier review errors on my part.

PR #47, opened 2022-07-15
Update `fugu-apis.js`

tomayac:
I have a generator script over in the old legacy repo: https://github.com/HTTPArchive/legacy.httparchive.org/blob/master/bin/create-fugu.js. For now, I have just copied its output manually over here. Would it be fine for me to add the generator script to the present repo?


rviscomi:
Yeah it'd be fine to migrate that script over here.

tunetheweb:
> Would it be fine for me to add the generator script to the present repo?

Think so but not into the `dist` folder as that’s automatically copied to WPT agents and run. FYI @pmeenan

tomayac:
Added the script. If more metrics need a generator script, we could generalize the `npm prepare` script. For now, it just executes my script.

rviscomi (approved):
Linter is complaining but otherwise LGTM

tomayac:
Made the linter happy 😃.

PR #52, opened 2022-09-28
Added null-check to body when checking JavaScript sourcemaps

### LINKED ISSUE #51: javascript metric in many cases empty
Hi,

I have noticed that the javascript custom metric get an improvement in https://github.com/HTTPArchive/custom-metrics/pull/42 , which now also collect statistics over the server generated HTML. 
That's really cool since that was something we have missed beforehand. 

However, it seems that the JavaScript metric is, for example, on the url `https://www.decathlon.de/` just empty. 
```
SELECT
  JSON_EXTRACT_SCALAR(payload, '$._javascript') AS javascript,
FROM
  `httparchive.pages.2022_06_01_*`
where url like 'https://www.decathlon.de%'  
limit 1
```

If have quickly validated that finding, and it looks like that this not unique to that url.
```
SELECT
  count(*) as records,
  count(JSON_EXTRACT_SCALAR(payload, '$._javascript')) AS javascript,
FROM
  `httparchive.pages.2022_06_01_*`
```

Shows that the metric is on ~2M cases null . 

Also, the _wpt_bodies metric reports some errors in that cases. 

```
{
  "log": [
    {
      "type": "error",
      "context": "anchors",
      "exception": {},
      "message": "Cannot read properties of null (reading 'length')"
    },
    {
      "type": "error",
      "context": "title",
      "exception": {},
      "message": "Cannot read properties of null (reading 'length')"
    },
    {
      "type": "error",
      "context": "meta_description",
      "exception": {},
      "message": "Cannot read properties of null (reading 'length')"
    },
    {
      "type": "error",
      "context": "hreflangs",
      "exception": {},
      "message": "Cannot read properties of null (reading 'length')"
    },
    {
      "type": "error",
      "context": "headings",
      "exception": {},
      "message": "Cannot read properties of null (reading 'length')"
    },
    {
      "type": "error",
      "context": "structured_data",
      "exception": {},
      "message": "Cannot read properties of null (reading 'length')"
    },
    {
      "type": "error",
      "context": "raw_html",
      "exception": {},
      "message": "Cannot read properties of null (reading 'length')"
    },
    {
      "type": "error",
      "context": "canonicals",
      "exception": {},
      "message": "Cannot read properties of null (reading '0')"
    },
    {
      "type": "error",
      "context": "robots",
      "exception": {},
      "message": "Cannot read properties of null (reading '0')"
    },
    {
      "type": "error",
      "context": "visible_words",
      "exception": {},
      "message": "Cannot read properties of null (reading 'length')"
    },
    {
      "type": "error",
      "context": "general",
      "message": "Failed to create the custom metrics object: Cannot read properties of null (reading 'length')",
      "exception": {}
    }
  ]
}
```

It looks like that the underlying issue is that the magic $WPT_REQUESTS is maybe null? 

We see a similar issue on several other domains where we are 
working with. All of them are quite large E-Commerce companies in Germany / EU. Not sure but that can of cause bias statistics which you may collect over that metrics.

It would be cool if someone can take a look into this or maybe give me a pointer how I can help to find the underlying issue, 

Thanks in advance

kevinfarrugia:
When the script response does not include a body, for example in the case of a 301 redirect, the custom metric is throwing an unhandled exception and returning `null`.

**Test**
See custom metric `__debug_javascript` on the following URL: https://webpagetest.httparchive.org/result/220928_FZ_D/1/details/#waterfall_view_step1

The test includes a 301 redirect for the script file https://fast.smarketer.de/api/js/fast.js .

**Actual**

Currently this was returning `null`. See the custom metric `__debug_javascript` on the following URL which uses the custom metric prior to the fix: https://webpagetest.httparchive.org/result/220928_70_E/1/details/#waterfall_view_step1

Closes #51 .

### DIFF
--- a/dist/javascript.js
+++ b/dist/javascript.js
@@ -39,7 +39,7 @@ let getSourceMaps = (async () => {
   let sourcemapURLs = scripts
     .map((n) => {
       if (n) {
-        let url = n.body.match(sourcemapRegex)?.[1];
+        let url = n.body?.match(sourcemapRegex)?.[1];
 
         if (url) {
           // Source map URL is relative to stylesheet URL


tunetheweb:
LGTM, but will ask @pmeenan to see if he’s happy with this before merging as he had a different theory.

pmeenan:
My JS foo is middling at best. If an exception during a filter call doesn't
cause issues then my theory doesn't apply.

Either way, this is at least one cause of issues so merge away.

kevinfarrugia:
if `type` is undefined or null it would return false and that item would not be picked up by the filter so I do not think it is an issue.

pmeenan (approved):

PR #57, opened 2022-11-03
Added setter to observed properties

### LINKED ISSUE #54: Re-enable the observer custom metric
We had to disable the observer custom metric because it was interfering with the page rendering in some cases.

Investigate how that interference happened and come up with a workaround so that we can re-enable this feature.

kevinfarrugia:
Fixes #54


inject-dist/observers.js
```diff
@@ -172,6 +172,9 @@ function initializeObserver(pathname) {
 
         // Return the original feature.
         return original;
+      },
+      set: () => {
+        return original;
```
rviscomi, line 177: If an application attempts to overwrite one of these properties, would this prevent it from working, since it's not actually setting the new value?

Worth testing if this works:

```suggestion
      set: (new) => {
        original = new;
```

rviscomi (commented):
Could you take a look at the merge conflict?

inject-dist/observers.js
kevinfarrugia: The above gives a number of issues (`undefined` or `null`), while my proposed solution removes any setters which are added through the app's code. So neither is a good solution as yet. I will try investigate the `null`/`undefined` to understand what's going on .

In the meantime I tried to rewrite it using a `proxy` but we are unable to proxy all objects. For example, it is not allowing me to proxy `window.navigator`.

kevinfarrugia:
I will close this and re-open when I have a working & tested version. I am not confident that the proposed solution works.

PR #62, opened 2023-01-11
CMS custom metric

rviscomi:
- Boolean check for whether the WordPress block theme was detected


felixarntz (approved):
Thanks @rviscomi, LGTM!

PR #64, opened 2023-01-14
Implement custom metrics to count stylesheets that are within vs after the `<head>` tag

felixarntz:
We would like to get a better idea on how often stylesheets are used in `<head>` vs in `<body>`, and we also would like to be able to assess how common certain combinations are.

For example, we could assume (without 100% reliability) that sites that inline critical CSS have the following:
* no external CSS in `<head>`
* at least 1 inline CSS in `<head>`
* at least 1 external CSS in `<body>`

The custom metrics here should allow us to query for that.

### DIFF
--- a/dist/css.js
+++ b/dist/css.js
@@ -14,12 +14,12 @@ function countInlineCssInHead() {
   return document.querySelectorAll( 'head style' ).length;
 }
 
-function countExternalCssAfterHead() {
-  return document.querySelectorAll( 'link[rel="stylesheet"]' ).length - countExternalCssInHead();
+function countExternalCssInBody() {
+  return document.querySelectorAll( 'body link[rel="stylesheet"]' ).length;
 }
 
-function countInlineCssAfterHead() {
-  return document.querySelectorAll( 'style' ).length - countInlineCssInHead();
+function countInlineCssInBody() {
+  return document.querySelectorAll( 'body style' ).length;
 }
 
 return JSON.stringify({
@@ -72,7 +72,7 @@ return JSON.stringify({
     ).length > 0,
 
   externalCssInHead: countExternalCssInHead(),
-  externalCssAfterHead: countExternalCssAfterHead(),
+  externalCssInBody: countExternalCssInBody(),
   inlineCssInHead: countInlineCssInHead(),
-  inlineCssAfterHead: countInlineCssAfterHead(),
+  inlineCssInBody: countInlineCssInBody(),
 });
--- a/dist/css.js
+++ b/dist/css.js
@@ -6,6 +6,22 @@ const PREFERS_COLOR_SCHEME_REGEXP =
 
 const bodies = $WPT_BODIES;
 
+function countExternalCssInHead() {
+  return document.querySelectorAll( 'head link[rel="stylesheet"]' ).length;
+}
+
+function countInlineCssInHead() {
+  return document.querySelectorAll( 'head style' ).length;
+}
+
+function countExternalCssAfterHead() {
+  return document.querySelectorAll( 'link[rel="stylesheet"]' ).length - countExternalCssInHead();
+}
+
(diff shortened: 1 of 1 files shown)


felixarntz:
@rviscomi Would you mind having a look?

cc @kt-12 for visibility: If/once this gets merged, it'll eventually become available in HTTP Archive as a metric we can easily query for.

dist/css.js
```diff
@@ -6,6 +6,22 @@ const PREFERS_COLOR_SCHEME_REGEXP =
...
+function countExternalCssInHead() {
+  return document.querySelectorAll( 'head link[rel="stylesheet"]' ).length;
+}
+
+function countInlineCssInHead() {
+  return document.querySelectorAll( 'head style' ).length;
+}
+
+function countExternalCssAfterHead() {
+  return document.querySelectorAll( 'link[rel="stylesheet"]' ).length - countExternalCssInHead();
```
tunetheweb, line 18: Why not just use the same as above but with `body`? In which case should also change the function name.

```suggestion
function countExternalCssInBody() {
  return document.querySelectorAll( 'body link[rel="stylesheet"]' ).length;
```

dist/css.js
```diff
@@ -6,6 +6,22 @@ const PREFERS_COLOR_SCHEME_REGEXP =
...
+}
+
+function countInlineCssAfterHead() {
+  return document.querySelectorAll( 'style' ).length - countInlineCssInHead();
```
tunetheweb, line 22: Ditto

```suggestion
function countInlineCssInBody() {
  return document.querySelectorAll( 'body style' ).length;
```

dist/css.js
```diff
@@ -54,4 +70,9 @@ return JSON.stringify({
     document.querySelectorAll(
       'link[rel="stylesheet"][media*="prefers-color-scheme"]',
     ).length > 0,
+
+  externalCssInHead: countExternalCssInHead(),
+  externalCssAfterHead: countExternalCssAfterHead(),
+  inlineCssInHead: countInlineCssInHead(),
+  inlineCssAfterHead: countInlineCssAfterHead(),
```
tunetheweb, line 77:
```suggestion
  externalCssInHead: countExternalCssInHead(),
  externalCssAfterHead: countExternalCssInBody(),
  inlineCssInHead: countInlineCssInHead(),
  inlineCssAfterHead: countInlineCssInBody(),
```

Though I'm not sure this is worth a separate function since it's pretty simple. The example above `css_in_js` inlines this.

dist/css.js
felixarntz: Good point, fixed in https://github.com/HTTPArchive/custom-metrics/pull/64/commits/70a1cda990ffc87c50bca29342656b9690dda48e.

dist/css.js
felixarntz: See above, fixed in https://github.com/HTTPArchive/custom-metrics/pull/64/commits/70a1cda990ffc87c50bca29342656b9690dda48e

dist/css.js
felixarntz: IMO separate functions are generally easier to read and provide better overview than inline, but if you prefer it inline I can change it.

tunetheweb: I could go either way. Happy to leave as is.

tunetheweb (approved):
Change LGTM but let's let @rviscomi review before we merge as he knows these better.

Any chance of a few WebPageTest example run links to make sure it's working as intended?
Details here: https://github.com/HTTPArchive/custom-metrics#testing

felixarntz:
@tunetheweb Thanks for the hint, I gave it a test in https://www.webpagetest.org/details.php?test=230116_BiDcBG_95V&run=1&cached=0 (testing https://wordpress.org/). Looks like it's working as expected.

rviscomi (approved):

PR #66, opened 2023-01-19
add custom metrics for Aurora

### LINKED ISSUE #59: Create a custom metric for Aurora metrics
The Aurora team works a lot with JS frameworks and a unique challenge they face is working with client-side rendered apps. These pages are much harder to analyze in HTTP Archive because the HTML response bodies contain empty app shells. Creating a new custom metric for Aurora will help them generate new metrics like feature adoption based on the fully rendered page.

kara:
This commit adds a custom metric for detecting
when the Angular image directive is used and
whether it is marked as "priority". This was
previously difficult to detect using HTTP
Archive because most Angular apps are client-side
rendered, so the output would not be available
in the normal response body table. Closes #59.

WPT test run: https://www.webpagetest.org/result/230119_AiDc4H_H6K/2/details/

### DIFF
--- a/dist/aurora_apis.js
+++ b/dist/aurora_apis.js
@@ -11,7 +11,7 @@ function getAngularImagePriorityCount() {
     return document.querySelectorAll('img[ng-img][priority]').length;
 }
 
-return JSON.stringify({
+return {
     ng_img_user: isAngularImageDirUser(),
     ng_priority_img_count: getAngularImagePriorityCount()
-});
+};
--- a/dist/aurora_apis.js
+++ b/dist/aurora_apis.js
@@ -1,0 +1,17 @@
+//[aurora]
+// Uncomment the previous line for testing on webpagetest.org
+
+// Detects if NgOptimizedImage is in use on the page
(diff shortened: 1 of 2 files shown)


dist/aurora_apis.js
```diff
@@ -0,0 +1,17 @@
...
+
+// Detects count of NgOptimizedImage instances with a priority attr 
+function getAngularImagePriorityCount() {
+    return document.querySelectorAll('img[ng-img][priority]').length;
+}
+
+return JSON.stringify({
+    ng_img_user: isAngularImageDirUser(),
+    ng_priority_img_count: getAngularImagePriorityCount()
+});
```
rviscomi, line 17: Older metrics required scalar outputs, but WPT can handle anything serializable including objects. This also simplifies how we query the results:

```suggestion
return {
    ng_img_user: isAngularImageDirUser(),
    ng_priority_img_count: getAngularImagePriorityCount()
};
```

rviscomi (changes requested):

rviscomi (approved):

PR #67, opened 2023-01-24
Better handle redirected pages

pmeenan:
This considers both the original URL that WPT navigated to when testing as well as the current document's URL when crawling links to reduce cases where redirects are crawled twice.

Fix https://github.com/HTTPArchive/data-pipeline/issues/139

Sample test result: https://webpagetest.httparchive.org/result/230124_GN_1/

newlinks (with the change):
```json
[
	"https:\/\/almanac.httparchive.org\/en\/2022\/table-of-contents",
	"https:\/\/almanac.httparchive.org\/en\/2022\/methodology",
	"https:\/\/almanac.httparchive.org\/en\/2022\/contributors",
	"https:\/\/almanac.httparchive.org\/en\/2022\/seo",
	"https:\/\/almanac.httparchive.org\/en\/accessibility-statement",
	"https:\/\/almanac.httparchive.org\/en\/rss.xml"
]
```

crawl_links (current metric)
```json
[
	"https:\/\/almanac.httparchive.org\/en\/2022\/",
	"https:\/\/almanac.httparchive.org\/en\/2022\/table-of-contents",
	"https:\/\/almanac.httparchive.org\/en\/2022\/methodology",
	"https:\/\/almanac.httparchive.org\/en\/2022\/contributors",
	"https:\/\/almanac.httparchive.org\/en\/2022\/seo",
	"https:\/\/almanac.httparchive.org\/en\/2022\/#maincontent",
	"https:\/\/almanac.httparchive.org\/en\/accessibility-statement",
	"https:\/\/almanac.httparchive.org\/en\/rss.xml"
]
```

### DIFF
--- a/dist/crawl_links.js
+++ b/dist/crawl_links.js
@@ -41,6 +41,7 @@ const getLinks = function(visibleOnly){
             if (rect.width > 1 && rect.height > 1 && (is_visible || !visibleOnly) && e) {
                 const url = e.href;
                 const baseUrl = url.split('#')[0]
+                // Link is not the current page.
                 if (baseUrl != testUrl && baseUrl != currentUrl) {
                     if (sameOrigin(url, testUrl) || sameOrigin(url, currentUrl)) {
                         let size = rect.width * rect.height;
--- a/dist/crawl_links.js
+++ b/dist/crawl_links.js
@@ -37,6 +37,7 @@ const getLinks = function(visibleOnly){
                                 style.visibility !== 'hidden' &&
                                 style.display !== 'none' &&
                                 intersectRect(rect, viewport);
+            // Link is user-visible.
             if (rect.width > 1 && rect.height > 1 && (is_visible || !visibleOnly) && e) {
                 const url = e.href;
                 const baseUrl = url.split('#')[0]
--- a/dist/crawl_links.js
+++ b/dist/crawl_links.js
@@ -39,10 +39,9 @@ const getLinks = function(visibleOnly){
                                 intersectRect(rect, viewport);
             // Link is user-visible.
             if (rect.width > 1 && rect.height > 1 && (is_visible || !visibleOnly) && e) {
-                const url = e.href;
-                const baseUrl = url.split('#')[0]
+                const url = e.href.split('#')[0];
                 // Link is not the current page.
-                if (baseUrl != testUrl && baseUrl != currentUrl) {
+                if (url != testUrl && url != currentUrl) {
                     // Link is to the same origin as the current page.
                     if (sameOrigin(url, testUrl) || sameOrigin(url, currentUrl)) {
                         let size = rect.width * rect.height;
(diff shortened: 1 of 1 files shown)


pmeenan:
@rviscomi one thing that I was considering when looking at this is how we want to handle fragments.  Right now they are all considered separate urls and the test url to go to includes the fragment.  This could be useful for SPA deep navigations.

on the other hand, ignoring the fragment part would better account for a page that is linked to multiple times for different parts of the page by aggregating the size of all of the links. Probably a rounding error in the actual number of pages it impacts but worth considering if we should normalize when crawling.

dist/crawl_links.js
```diff
@@ -36,13 +37,17 @@ const getLinks = function(visibleOnly){
...
                                 style.display !== 'none' &&
                                 intersectRect(rect, viewport);
-            if (rect.width > 1 && rect.height > 1 && (is_visible || !visibleOnly) ) {
-                if (e && e.href.split('#')[0] != documentOrigin && sameOrigin(e.href, documentOrigin)) {
-                    let size = rect.width * rect.height;
-                    if (links[e.href] === undefined) {
-                        links[e.href] = size;
-                    } else {
-                        links[e.href] += size;
+            if (rect.width > 1 && rect.height > 1 && (is_visible || !visibleOnly) && e) {
```
rviscomi, line 40:
```suggestion
            // Link is user-visible.
            if (rect.width > 1 && rect.height > 1 && (is_visible || !visibleOnly) && e) {
```

dist/crawl_links.js
```diff
@@ -36,13 +37,17 @@ const getLinks = function(visibleOnly){
...
+                const url = e.href;
+                const baseUrl = url.split('#')[0]
+                if (baseUrl != testUrl && baseUrl != currentUrl) {
```
rviscomi, line 43:
```suggestion
                // Link is not the current page.
                if (baseUrl != testUrl && baseUrl != currentUrl) {
```

dist/crawl_links.js
```diff
@@ -36,13 +37,17 @@ const getLinks = function(visibleOnly){
...
+                    if (sameOrigin(url, testUrl) || sameOrigin(url, currentUrl)) {
```
rviscomi, line 44:
```suggestion
                    // Link is to the same origin as the current page.
                    if (sameOrigin(url, testUrl) || sameOrigin(url, currentUrl)) {
```

rviscomi (approved):
LGTM. Took me a minute to step through each of the conditions, so suggesting some comments to make that easier in the future.

pmeenan:
@rviscomi what do you think about fragments?  I'm inclined to remove the fragments and just crawl the pages (and count all links to the same page as the same).  That way we're also navigating to the page and not a position within a page.

pmeenan:
Decided that removing the fragments when crawling made the most sense so [added that change too](https://github.com/HTTPArchive/custom-metrics/pull/67/commits/dc747645364643779c21257de7b88dc79a872d4e).

PR #73, opened 2023-03-16
Add WordPress embed block detection

adamsilverstein:


dist/cms.js
```diff
@@ -5,9 +5,15 @@ function usesBlockTheme() {
...
+  has_embed_block: hasWordPressEmbedBlock()
 };
 
 return {
   wordpress
 };
+
+// Detects if a WordPress embed block is on the page
+function hasWordPressEmbedBlock() {
+  return !!document.querySelector('div.wp-block-embed');
```
rviscomi, line 18: One offline question about figure vs div, otherwise my only other comment would be to hoist this function higher in the file after `usesBlockTheme`

rviscomi (approved):

PR #74, opened 2023-03-17
add nuxt version detection for nuxt 3.3+

kara:
Nuxt 3.3.1 surfaces Nuxt version data. Through a
custom metric, we can now see Nuxt versions in
addition to Vue versions. This pattern is not
compatible with Wappalyzer, so a custom metric
is the only way to aggregate this data right
now.

Test run: https://www.webpagetest.org/result/230316_BiDcFV_C3D/1/details/

### DIFF
--- a/dist/aurora.js
+++ b/dist/aurora.js
@@ -27,7 +27,7 @@ function getVueVersionForNuxt() {
 
 // Detects Nuxt version directly for Nuxt 3.3 and later
 function getNuxtVersion() {
-  return __unctx__?.get('nuxt-app')?.use()?.versions.nuxt;
+  return __unctx__?.get('nuxt-app')?.use()?.versions?.nuxt;
 }
 
 // Detects React version (set earlier by inject_script)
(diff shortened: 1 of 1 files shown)


dist/aurora.js
```diff
@@ -18,13 +18,18 @@ function getAngularImagePriorityCount() {
...
+// Nuxt only exposes its own version for Nuxt apps on v3.3+
 function getVueVersionForNuxt() {
   const nuxt2Version = window.$nuxt?.$root?.constructor?.version;
   const nuxt3Version = document.querySelector('#__nuxt')?.__vue_app__?.version;
   return nuxt2Version || nuxt3Version;
 }
 
+// Detects Nuxt version directly for Nuxt 3.3 and later
+function getNuxtVersion() {
+  return __unctx__?.get('nuxt-app')?.use()?.versions.nuxt;
```
rviscomi, line 30: Can't hurt

```suggestion
  return __unctx__?.get('nuxt-app')?.use()?.versions?.nuxt;
```

rviscomi (approved):

dist/aurora.js
kara: Makes sense! Thanks for making the change!

PR #75, opened 2023-03-21
CMS: Add WordPress embed block counts by provider

adamsilverstein:
This adds counts for embed blocks used on WordPress pages, including a breakdown of the source of the embed.

For example, running `getWordPressEmbedBlockCounts ` on this test page: 

https://oembeds-test.instawp.xyz/test-many-oembeds/

returns this data:

image

Note: the page includes two embeds of _other pages on the same site_, hence the provider shown as `oembeds-test-instawp-xyz: 2 `

### DIFF
--- a/dist/cms.js
+++ b/dist/cms.js
@@ -9,9 +9,26 @@ function hasWordPressEmbedBlock() {
   return !!document.querySelector('figure.wp-block-embed');
 }
 
+// Count the number of WordPress embed blocks on the page, including a breakdown by type
+function getWordPressEmbedBlockCounts() {
+  const embedBlocks = document.querySelectorAll('figure.wp-block-embed');
+  const embedsByType = [];
+  for ( let embed of embedBlocks ) {
+    let provider = embed.className.split( ' ' ).pop().split( "-" ).pop();
+    embedsByType[ provider ] = embedsByType[ provider ] || 0;
+    embedsByType[ provider ]++;
+  }
+
+  return {
+    total: embedBlocks.length,
+    total_by_type: embedsByType
+  }
+}
+
 const wordpress = {
   block_theme: usesBlockTheme(),
-  has_embed_block: hasWordPressEmbedBlock()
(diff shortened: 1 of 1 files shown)


dist/cms.js
```diff
@@ -9,9 +9,36 @@ function hasWordPressEmbedBlock() {
...
+    total: embedBlocks.length,
+    total_by_type: embedsByType
+  }
+}
+
 const wordpress = {
   block_theme: usesBlockTheme(),
-  has_embed_block: hasWordPressEmbedBlock()
+  has_embed_block: hasWordPressEmbedBlock(),
+  embed_block_count: getWordPressEmbedBlockCounts()
```
adamsilverstein, line 41: @rviscomi is it ok to return the data with a nested array, or would it be better to put all of the provider counts and total count at the "top level" of the returned object?

rviscomi: Nested arrays/objects are totally fine

dist/cms.js
```diff
@@ -9,9 +9,36 @@ function hasWordPressEmbedBlock() {
...
 
+// Count the number of WordPress embed blocks on the page, including a breakdown by type
+function getWordPressEmbedBlockCounts() {
+  const embedBlocks = document.querySelectorAll('figure.wp-block-embed');
+  const embedsByType = [];
+  for (let embed of embedBlocks) {
+    let embedClasses = embed.className.split( ' ' );
+
+    // Find the provider classname, which starts with "is-provider-".
+    let provider = embedClasses.find( function( className ) {
```
rviscomi, line 20: This will return the first match. Is it ever possible for the same block to have multiple `is-provider` values?

adamsilverstein: the figure can only contain a single `is-provider-*` class, it is added on save here: https://github.com/WordPress/gutenberg/blob/74d1b0db4732f23085e788eb895f308b92c16aff/packages/block-library/src/embed/save.js#L24

rviscomi (approved):

PR #78, opened 2023-04-19
add error handling to avoid whole file breaks

kara:
In a recent PR, all Aurora metrics were unavailable on some pages because an individual metric threw
an error. This PR adds basic error handling so each metric is isolated and can't break the whole file.

Nuxt test run: https://www.webpagetest.org/result/230419_AiDcWV_CCS/2/details/
Angular test run: https://www.webpagetest.org/result/230419_BiDcXX_BXE/2/details/
React test run:  https://www.webpagetest.org/result/230419_BiDcH4_2K3/1/details/

Test run with typo in two functions (typo manually added in WPT):  https://www.webpagetest.org/result/230419_AiDcSC_CCW/1/details/

### DIFF
--- a/dist/aurora.js
+++ b/dist/aurora.js
@@ -1,6 +1,23 @@
 //[aurora]
 // Uncomment the previous line for testing on webpagetest.org
 
+// Wrap individual metric calls so errors don't break metrics in entire file.
+function runSafely(cb) {
+  try {
+    return cb() || null;
+  } catch (e) {
+    return logError(e);
+  }
+}
+
+function logError(e) {
+  try {
+    return `${e?.name}: ${e?.message}`;
+  } catch {
+    return null;
+  }
+}
+
 // Detects Angular version, which is added through runtime JS
 function getAngularVersion() {
   const versionEl = document.querySelector('[ng-version]');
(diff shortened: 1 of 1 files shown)


tunetheweb:
LGTM. You could make it slightly cleaner by moving the `|| null` into the function call.

kara:
> LGTM. You could make it slightly cleaner by moving the `|| null` into the function call.

Ah, great point. Updated! (updated the test run links as well)

tunetheweb (approved):

dist/aurora.js
```diff
@@ -1,6 +1,15 @@
 //[aurora]
 // Uncomment the previous line for testing on webpagetest.org
 
+// Wrap individual metric calls so errors don't break metrics in entire file.
+function runSafely(cb) {
+  try {
+    return cb() || null;
+  } catch {
+    return `ERROR`;
```
rviscomi, line 9: I'd usually just return null here so the results don't show up in the aggregate analysis. If you want to more clearly indicate that there was an error, maybe for debugging purposes, would including the exception message here be more helpful?

kara: Yeah, the "ERROR" was supposed to help me identify which URLs had issues, but including a snapshot of the error message would be even more helpful. Added it!

dist/aurora.js
```diff
@@ -1,6 +1,23 @@
...
+  try {
+    return cb() || null;
+  } catch (e) {
+    return logError(e);
+  }
+}
+
+function logError(e) {
+  try {
+    return `${e?.name}: ${e?.message}`;
```
rviscomi, line 15: A concrete prefix to anchor on, eg in case you need to filter out pages with errors.

```suggestion
    return `Error ${e?.name}: ${e?.message}`;
```

rviscomi (approved):
One last comment, then feel free to merge

dist/aurora.js
kara: Makes sense, committed your suggestion!

kara:
Thanks for the review! Should be good to merge now.

PR #79, opened 2023-05-25
Add custom metric for `html[lang]` attribute

swissspidy:
Extract the value of the `html[lang]` attribute to allow for easier comparison of sites with different languages.

Since country ≠ language, it's important to be able to detect the language information separately, and the `lang` attribute is the most useful indicator.

### DIFF
--- a/dist/markup.js
+++ b/dist/markup.js
@@ -208,7 +208,7 @@ try { // whole process is placed in a try/catch so we can log uncaught errors
     // Extract the <html lang=""> attribute value
     'lang': (() => {
       try {
-        return document.querySelector('html[lang]')?.getAttribute("lang") || null;
+        return document.querySelector('html[lang]')?.getAttribute("lang")
       }
       catch(e) {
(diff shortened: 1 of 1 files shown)


tunetheweb:
We already get this in [`almanac.js` where we get the full html node](https://github.com/HTTPArchive/custom-metrics/blob/8497c859ef0a7c99924981f369bb53eb3441bd6c/dist/almanac.js#L768) and all it's attributes.

You can see how to query it here: https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/sql/2022/markup/lang.sql

Or see results here: https://almanac.httparchive.org/en/2022/markup#fig-4
or for more details here: https://docs.google.com/spreadsheets/d/1grkd2_1xSV3jvNK6ucRQ0OL1HmGTsScHuwA8GZuRLHU/edit#gid=1194853402

swissspidy:
Ah, I missed this. Thanks! Will look into it.

PR #82, opened 2023-06-15
detect bound priority attr usages

kara:
Previously, the ng_priority_img_count custom metric could only detect unbound usages of the "priority" attribute because it was checking for it directly. With this change, we check the "fetchpriority"
attribute instead because that will also catch
cases where the attribute was bound (in which case it disappears from the DOM).

Test run: https://www.webpagetest.org/result/230615_BiDc8T_DS5/1/details/

### DIFF
--- a/dist/aurora.js
+++ b/dist/aurora.js
@@ -31,7 +31,7 @@ function isAngularImageDirUser() {
 
 // Detects count of NgOptimizedImage instances with a priority attr
 function getAngularImagePriorityCount() {
-    return document.querySelectorAll('img[ng-img][priority]').length;
+    return document.querySelectorAll('img[ng-img][fetchpriority=high]').length;
 }
 
 // Detects Vue version for Nuxt apps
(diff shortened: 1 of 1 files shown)


dist/aurora.js
```diff
@@ -31,7 +31,7 @@ function isAngularImageDirUser() {
 
 // Detects count of NgOptimizedImage instances with a priority attr
 function getAngularImagePriorityCount() {
-    return document.querySelectorAll('img[ng-img][priority]').length;
+    return document.querySelectorAll('img[ng-img][fetchpriority=high]').length;
```
rviscomi, line 34: Does it make sense to check for both?

```suggestion
    return document.querySelectorAll('img[ng-img]:is([priority],[fetchpriority=high])').length;
```

kara: Sure, I think it works either way (since "priority" should always set "fetchpriority=high"). But may as well be careful!

rviscomi (approved):

PR #83, opened 2023-06-26
Add fetchpriority attribute usage to Images custom metric

felixarntz:
This PR adds the `fetchpriority` attribute to the `Images` custom metric, allowing to monitor adoption and usage of the attribute and its values.

### DIFF
--- a/dist/Images.js
+++ b/dist/Images.js
@@ -18,6 +18,7 @@ var wptImages = function (win) {
             naturalHeight: el.naturalHeight,
             loading: el.getAttribute("loading"),
             decoding: el.getAttribute("decoding"),
+            fetchpriority: el.getAttribute("fetchpriority"),
             inViewport:
               el.getBoundingClientRect().bottom >= 0 &&
               el.getBoundingClientRect().right >= 0 &&


tunetheweb (approved):
LGTM. Note we already track this for LCP images in https://github.com/HTTPArchive/custom-metrics/blob/main/dist/performance.js (where we get ALL the attributes for the LCP element).

Example query: https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/sql/2022/performance/lcp_element_data_2.sql

felixarntz:
Thanks for the context @tunetheweb. Having it here too should help us monitor general usage, to e.g. also assess how many images get `fetchpriority="high"` that shouldn't.

tunetheweb:
Agreed! Do you happen to have a WebPageTest link showing this working? Can't see why it wouldn't but always good to have an example showing it definitely does.

felixarntz:
@tunetheweb I ran one here: https://www.webpagetest.org/result/230626_AiDcEE_C71/1/details/#:~:text=10.48-,IMAGES_WITH_FETCHPRIORITY,-%5B%7B%22url%22%3A%22https%3A%5C/%5C/web

PR #85, opened 2023-07-07
Add custom metrics for WordPress scripts

westonruter:
This PR gathers data for scripts that WordPress prints to a page via `WP_Scripts`. As of WordPress 6.3, first-party support is finally landing for [`async` and `defer` scripts](https://core.trac.wordpress.org/ticket/12009). This custom metric will help track adoption, including whether the intended strategy (`async` or `defer`) was not able to be applied due to a dependency or an `after` inline script. 

I made an [example plugin](https://gist.github.com/westonruter/a195bfd69dd8bc7f1007ba25acb8366f) that dumps out various scenarios into a console table:

image

I need to put that up on a test size to get the WebPageTest result.

### DIFF
--- a/dist/cms.js
+++ b/dist/cms.js
@@ -67,6 +67,7 @@ function getWordPressScripts() {
       async: script.async,
       defer: script.defer,
       intended_strategy: script.dataset.wpStrategy || null,
+      // For each external script, check if there is a related inline script
       after_script_size: getInlineScriptSize(document.getElementById(`${handle}-js-after`)),
       before_script_size: getInlineScriptSize(document.getElementById(`${handle}-js-before`)),
       extra_script_size: getInlineScriptSize(document.getElementById(`${handle}-js-extra`)),
--- a/dist/cms.js
+++ b/dist/cms.js
@@ -35,10 +35,52 @@ function getWordPressEmbedBlockCounts() {
   }
 }
 
+/**
+ * Obtains data about scripts that WordPress prints on the page.
+ *
+ * @returns {{}[]}
+ */
+function getWordPressScripts() {
+  const entries = [];
+
+  /**
+   * Checks whether the provided element is an inline script.
+   *
+   * @param {?Element} element Element to examine. May be null.
+   * @return {?number} Size of the inline script or null if the element isn't an inline script.
+   */
+  const getInlineScriptSize = (element) => {
+    if (element instanceof HTMLScriptElement && !element.src) {
+      return element.textContent.length;
+    }
+    return null;
+  };
+
+  const scripts = document.querySelectorAll('script[src][id$="-js"]');
+  for (const script of scripts) {
+    /** @var HTMLScriptElement script */
+    const handle = script.id.replace(/-js$/, '');
+    entries.push({
+      handle,
+      src: script.src,
+      in_footer: script.parentNode !== document.head,
+      async: script.async,
+      defer: script.defer,
+      intended_strategy: script.dataset.wpStrategy || null,
+      after_script_size: getInlineScriptSize(document.getElementById(`${handle}-js-after`)),
+      before_script_size: getInlineScriptSize(document.getElementById(`${handle}-js-before`)),
+      extra_script_size: getInlineScriptSize(document.getElementById(`${handle}-js-extra`)),
+      translations_script_size: getInlineScriptSize(document.getElementById(`${handle}-js-translations`)),
+    });
+  }
+  return entries;
+}
+
 const wordpress = {
   block_theme: usesBlockTheme(),
   has_embed_block: hasWordPressEmbedBlock(),
-  embed_block_count: getWordPressEmbedBlockCounts()
+  embed_block_count: getWordPressEmbedBlockCounts(),
+  scripts: getWordPressScripts(),
 };
 
 return {
--- a/dist/cms.js
+++ b/dist/cms.js
@@ -44,10 +44,10 @@ function getWordPressScripts() {
   const entries = [];
 
   /**
-   * Checks whether the provided element is an inline script.
+   * Returns the number of characters in an inline script.
+   * Returns null for non-inlined scripts
    *
    * @param {?Element} element Element to examine. May be null.
-   * @return {?number} Size of the inline script or null if the element isn't an inline script.
    */
   const getInlineScriptSize = (element) => {
     if (element instanceof HTMLScriptElement && !element.src) {


felixarntz (approved):
Thanks @westonruter, this looks great to me and has everything we need. It would be great to have a WebPageTest example for demonstration purposes.

felixarntz:
@tunetheweb Could you review this please when you get a chance?

westonruter:
Here are the custom metrics captured in a WPT example: https://www.webpagetest.org/result/230707_AiDcEF_F8W/1/details/#:~:text=F%22%2C%22securityHeadersScore%22%3A0%7D-,WORDPRESSSCRIPTS,-%5B%7B%22handle%22%3A%22wp%2Dblock

felixarntz:
@westonruter Thanks! Looking at the WPT result, while I assume this works as expected, none of the results have an intended strategy - likely because the site is running WP 6.2.2 rather than 6.3 Beta? That would explain it but makes it somewhat unreliable as a test case for your demo plugin :)

westonruter:
@felixarntz Right you are! I just updated to 6.3-beta3. Here's an updated WPT run: https://www.webpagetest.org/result/230707_AiDcVA_FXV/1/details/#:~:text=F%22%2C%22securityHeadersScore%22%3A0%7D-,WORDPRESSSCRIPTS,-%5B%7B%22handle%22%3A%22wp%2Dblock

dist/cms.js
```diff
@@ -35,10 +35,52 @@ function getWordPressEmbedBlockCounts() {
...
+ */
+function getWordPressScripts() {
+  const entries = [];
+
+  /**
+   * Checks whether the provided element is an inline script.
+   *
+   * @param {?Element} element Element to examine. May be null.
+   * @return {?number} Size of the inline script or null if the element isn't an inline script.
+   */
```
tunetheweb, line 51: This isn't an accurate comment.

```suggestion
  /**
   * Returns the number of characters in an inline script.
   * Returns null for non-inlined scripts
   *
   * @param {?Element} element Element to examine. May be null.
   */
```

dist/cms.js
```diff
@@ -35,10 +35,52 @@ function getWordPressEmbedBlockCounts() {
...
+  };
+
+  const scripts = document.querySelectorAll('script[src][id$="-js"]');
+  for (const script of scripts) {
+    /** @var HTMLScriptElement script */
+    const handle = script.id.replace(/-js$/, '');
+    entries.push({
+      handle,
+      src: script.src,
+      in_footer: script.parentNode !== document.head,
```
tunetheweb, line 66: This doesn't imply it's in the footer. Just that it's in the body. Should the name be changed?

```suggestion
      in_body: script.parentNode !== document.head,
```

dist/cms.js
```diff
@@ -35,10 +35,52 @@ function getWordPressEmbedBlockCounts() {
...
+      async: script.async,
+      defer: script.defer,
+      intended_strategy: script.dataset.wpStrategy || null,
+      after_script_size: getInlineScriptSize(document.getElementById(`${handle}-js-after`)),
+      before_script_size: getInlineScriptSize(document.getElementById(`${handle}-js-before`)),
+      extra_script_size: getInlineScriptSize(document.getElementById(`${handle}-js-extra`)),
+      translations_script_size: getInlineScriptSize(document.getElementById(`${handle}-js-translations`)),
```
tunetheweb, line 73: I think it's worth adding a comment here as I was a bit confused here.

```suggestion
      // For each external script, check if there is a related inline script
      after_script_size: getInlineScriptSize(document.getElementById(`${handle}-js-after`)),
      before_script_size: getInlineScriptSize(document.getElementById(`${handle}-js-before`)),
      extra_script_size: getInlineScriptSize(document.getElementById(`${handle}-js-extra`)),
      translations_script_size: getInlineScriptSize(document.getElementById(`${handle}-js-translations`)),
```

dist/cms.js
```diff
@@ -35,10 +35,52 @@ function getWordPressEmbedBlockCounts() {
...
+   * @return {?number} Size of the inline script or null if the element isn't an inline script.
+   */
+  const getInlineScriptSize = (element) => {
+    if (element instanceof HTMLScriptElement && !element.src) {
+      return element.textContent.length;
+    }
+    return null;
+  };
+
+  const scripts = document.querySelectorAll('script[src][id$="-js"]');
```
tunetheweb, line 59: Will all scripts have an id that ends with `-js` in WordPress?

Also this will pick up scripts on other sites that conform to this standard even if they aren't WordPress sites, but think that's OK. Unless there's an easy way to detect if this is a WordPress site and limit this custom metric to just that?

tunetheweb (changes requested):
Looks mostly good, but have a couple of questions and suggested a few tweaks

dist/cms.js
westonruter: Good catch. I wrote this initially when it was just checking if it was present. Thank you.

dist/cms.js
westonruter: The reason for using `in_footer` is that this is the parameter name used in the WordPress [`wp_register_script`](https://developer.wordpress.org/reference/functions/wp_register_script/)/[`wp_enqueue_script`](https://developer.wordpress.org/reference/functions/wp_enqueue_script/) APIs. It is in the head if `script.parentNode === document.head`, so the opposite of that `!==` means it is `in_footer`.

dist/cms.js
westonruter: > Will all scripts have an id that ends with `-js` in WordPress?

Yes, all WordPress scripts get the `-js` suffix (cf. [source code](https://github.com/WordPress/wordpress-develop/blob/f933aa3616aa8d49bdd67886e4af287579093c25/src/wp-includes/class-wp-scripts.php#L414)).

> Also this will pick up scripts on other sites that conform to this standard even if they aren't WordPress sites, but think that's OK. Unless there's an easy way to detect if this is a WordPress site and limit this custom metric to just that?

That's true. There are various ways of sniffing for whether a site is WordPress, as can be [seen in Wappalyzer](https://github.com/wappalyzer/wappalyzer/blob/e27056248a7b9c678e5ff075558ba7b2611d434d/src/technologies/w.json#L1939-L1978). I don't suppose this `cms.js` script has access to the Wappalyzer test results? Otherwise, I'm not sure it would be worthwhile from a maintenance POV to copy the detection logic into this script.

dist/cms.js
tunetheweb: OK. Though this script could also pick up other scripts not inserted by that. Anyway, happy enough to go with `in_footer` anyway since it's in a `wordpress` block.

tunetheweb (approved):

dist/cms.js
tunetheweb: Yeah let's leave it as is.

PR #86, opened 2023-07-07
add svelte version detection

kara:
Svelte recently added major version detection as of v4. It's exposed in a way that isn't supported by Wappalyzer, so adding a custom metric.


dist/aurora.js
```diff
@@ -52,11 +52,17 @@ function getReactVersion() {
   return window.react_renderer_version;
 }
 
+// Detects Svelte version
+function getSvelteVersion() {
+  return window.__svelte?.v ? Array.from(window.__svelte?.v)[0] : null;
```
tunetheweb, line 57: `toString()` rather than `[0]` would allow you to return all values in case there's multiple versions of Svelte on a page (I'm not sure if that is possible?). It would still return the same in the common casewhen there is only a singe version.

```suggestion
  return window.__svelte?.v ? Array.from(window.__svelte?.v).toString() : null;
```

tunetheweb (approved):
Looks good. One question/suggestion.

dist/aurora.js
kara: The short answer is that it's not yet possible (only one major version supported right now), but we may as well make it future-proof! Committed your suggestion

tunetheweb (approved):

PR #87, opened 2023-07-11
Fix bug in WordPress embed block count

tunetheweb:
Should be an object, not an array

[Old test](https://www.webpagetest.org/result/230711_BiDcCV_DZW/1/details/#waterfall_view_step1) fails:

**CMS**
```json
{"wordpress":{"block_theme":false,"has_embed_block":true,"embed_block_count":{"total":9,"total_by_type":[]}}}
```

[New Test](https://www.webpagetest.org/result/230711_BiDcED_E10/1/details/#waterfall_view_step1) succeeds:

**CMS**
```json
{"wordpress":{"block_theme":false,"has_embed_block":true,"embed_block_count":{"total":9,"total_by_type":{"youtube":9}}}}
```


adamsilverstein (approved):
Looks good to me, thanks for fixing this!

PR #89, opened 2023-07-24
Github action for custom metric tests with WPT API

### LINKED ISSUE #6: Integrate WPT GitHub action
https://github.com/WPO-Foundation/webpagetest-github-action

When someone creates a PR to change a custom metric, the GH action should automatically test it on WPT and comment on the PR with the output of the affected custom metric(s) and a link to the results.

It would also be nice to have a way to tell the GH action which URL(s) to use for testing and whether to use desktop and/or mobile, otherwise there should be sensible defaults. For example, the action could search the PR description for URLs and test config options like desktop/mobile.

max-ostapenko:
Resolves #6 

As pointed in [README](https://github.com/HTTPArchive/custom-metrics#testing), each custom metric adjustment can be tested with WPT.

PR includes:
- [x] test workflow,
- [x] WPT runner script,
- [x] unit tests,
- [x] parsing custom websites to test from PR body (see PR template).

---
**Test websites**:

- https://example.com/


rviscomi:
Oh wow thank you for picking this up!

.github/workflows/wpt-test.yml
```diff
@@ -0,0 +1,36 @@
...
+            if [ -f "tests/$TEST_NAME.test.js" ]
+            then
+              npm run test -- $TEST_NAME
+            else
+              echo "There are no test cases for $TEST_NAME"
+              exit 1
+            fi
+          done
+        env:
+          WPT_API_KEY: ${{ secrets.WPT_API_KEY }}
```
max-ostapenko, line 36: @rviscomi could you add a WPT secret as expected here

max-ostapenko: And if you could share one to speed up local dev - would be great!

rviscomi: Set up the `WPT_API_KEY` repo secret and DM'd you in the HTTP Archive slack with a private key you can use for testing.

max-ostapenko:
Unfortunately WebPageTest/webpagetest-api doesn't support POST requests to WPT API.
So when we try to attach custom metrics code (privacy.js ~ 14Kb) - test run request fails.

I don't expect [WPT issue](https://github.com/WebPageTest/webpagetest-api/issues/177) to be resolved soon.
So I think of launching first version with limited size support only. And later introduce support of POST requests.

@rviscomi @pmeenan do you have alternative ideas/comments?

pmeenan:
Could also post to the API directly. Though it's more work to implement the
polling. Or fork the project and implement post (might be easier).

max-ostapenko:
Thanks @pmeenan
I think I see the easy way to make a fork and probably even make a PR to WPT.

max-ostapenko:
I tested API as follows:
```bash
curl --location 'https://www.webpagetest.org/runtest.php?url=https%3A%2F%2Fexample.com%2F&f=json&k=INSERT_WPT_KEY' \
--header 'Content-Type: application/json' \
--data '{
    "location": "Dulles:Chrome.Native",
    "mobile": 1
}'
```
but still got `mobile: 0` and `connectivity: Cable` in the test results.

@pmeenan Did I misunderstood the doc and [runtest.php](https://www.webpagetest.org/runtest.php) doesn't look at parameters passed in POST request body, just URL query?
OR am I doing something wrong?

.github/workflows/wpt-test.yml
```diff
@@ -0,0 +1,45 @@
...
+      - name: Run WebPageTest
+        run: |
+          EXPECTED_TESTS=$(for file in $(git diff --name-only --diff-filter=ACMRT ${{ github.event.pull_request.base.sha }} \
+            ${{ github.event.pull_request.head.sha }} | grep -E "^dist/.*\.js$"); do basename "$file"; done | cut -d\. -f1 | sort | uniq)
+          CUSTOM_METRIC_FILES=("00_reset", "a11y" "almanac" "aurora" "cms" "css-variables" "css" \
+            "ecommerce" "example" "fugu-apis" "javascript" "markup" "media" "observers" "origin-trials" \
+            "parsed_css" "performance" "privacy" "pwa" "responsive_images" "robots_meta" "robots_txt" \
+            "sass" "security" "structured-data" "third-parties" "valid-head" "well-known" "wpt_bodies")
+
+          # These tests are too big to run using WPT API (custom metrics code needs to be less than 8Kb)
```
max-ostapenko, line 27: GET request URL accepted by WPT server is around 6Kb long.
We can start with limited support and see how helpful the tests are.

max-ostapenko:
@pmeenan Maybe there is some option to use our custom WPT instance to run the tests with complete custom metrics code included?
I mean it works somehow in PROD crawling.

Otherwise, @rviscomi @pmeenan, looking forward to your review.

rviscomi:
@pmeenan WDYT about using the private instance for testing? My only concern would be somehow interfering with the crawl queue.

pmeenan:
Sorry for the slow replies. Came down with covid after IETF :(

The private instance has a single test machine and is independent of the
crawl. I'd just need to assign a key.

As far as the API goes, POST should work as long as it is multipart mime.
And not mixed with query params and post fields.

max-ostapenko:
[WPT API library PR](https://github.com/WebPageTest/webpagetest-api/pull/178) took quite some time, but I hope it will be merged into original wrapper.

For now we'll use a forked version.

max-ostapenko:
Now you could even include all custom metrics in a single test case.
But by default - only chapter related to keep it fast.

tests/wpt.js
```diff
@@ -0,0 +1,38 @@
+const WebPageTest = require('webpagetest');
+const fs = require('fs');
+
+const wptServer = 'www.webpagetest.org';
+let wptApiKey = process.env.WPT_API_KEY;
```
rviscomi, line 5:
```suggestion
const wptApiKey = process.env.WPT_API_KEY;
```

tests/wpt.js
```diff
@@ -0,0 +1,38 @@
...
+  return new Promise((resolve, reject) => {
+    wpt.runTestAndWait(url, options, (error, response) => {
+      console.log(`Running ${options.label} tests for ${url}`);
+      if (error || response.statusCode !== 200) {
+        console.error(`Test for ${url} failed`);
+        console.error(error || response);
+        reject(error || response);
+      } else {
+        console.log(`Test for ${url} succeeded`);
+        console.log(response);
```
rviscomi, line 29: The response could be huge, not sure if it's helpful to log the whole thing. Maybe just the test ID:

```suggestion
        console.log(`Test ID: ${JSON.parse(response.data.id)}`);
```

tests/wpt.js
```diff
@@ -0,0 +1,38 @@
+const WebPageTest = require('webpagetest');
+const fs = require('fs');
+
+const wptServer = 'www.webpagetest.org';
+let wptApiKey = process.env.WPT_API_KEY;
+let wpt = new WebPageTest(wptServer, wptApiKey);
```
rviscomi, line 6:
```suggestion
const wpt = new WebPageTest(wptServer, wptApiKey);
```

rviscomi (changes requested):

dist/privacy.test.js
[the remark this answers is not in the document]
max-ostapenko: Moved.
Where is the script to integrate metrics for the prod tests?
I think there are quite some things that could be reused.

.github/workflows/wpt-test.yml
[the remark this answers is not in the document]
max-ostapenko: These are available in action logs.
My idea was that it's better to keep PR clean from test logs, because it not helpful for review discussion.

tests/wpt.js
max-ostapenko: I kept the test results URL.

.github/workflows/wpt-test.yml
```diff
@@ -0,0 +1,40 @@
...
+
+      - name: Install dependencies
+        run: npm install webpagetest && npm install jest
+
+      - name: Run WebPageTest
+        run: |
+          EXPECTED_TESTS=$(for file in $(git diff --name-only --diff-filter=ACMRT \
+            ${{ github.event.pull_request.base.sha }} ${{ github.event.pull_request.head.sha }} | \
+            grep -E "^dist/.*\.js$"); do basename "$file"; done | cut -d\. -f1 | sort | uniq)
+          CUSTOM_METRIC_FILES=("00_reset", "a11y" "almanac" "aurora" "cms" "css-variables" "css" "ecommerce" \
```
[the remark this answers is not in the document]
max-ostapenko: Oh, I think I misunderstood the files.
Now I see that each file is integrated as a custom metrics (has a return statement in it).

I removed the list. If there are any exceptions - let me know.

.github/workflows/wpt-test.yml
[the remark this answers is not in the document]
rviscomi: We used to manually put test results in the PR comments ([example](https://github.com/HTTPArchive/custom-metrics/pull/24)) as a way of helping reviewers confirm that the changes are working working and see what the expected output is. For me at least, I think it's helpful. But let me know if it's too much work.

dist/privacy.test.js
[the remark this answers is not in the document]
rviscomi: Somewhere in the core WebPageTest code, maybe @pmeenan can point you to it.

tests/privacy.test.js
rviscomi, line 1: Now that these are in the `tests/` directory, do we need the `.test.js` suffix or can we simplify to `.js`?

tests/privacy.test.js
```diff
@@ -0,0 +1,43 @@
...
+const path = require('path');
+const options = {
+  label: path.basename(__filename).replace('.test.js', '')
+}
+
+beforeAll(async () => {
+  theverge = await runWPTTest("https://www.theverge.com/", options);
+  rtl = await runWPTTest("https://www.rtl.de/", options);
+  educations = await runWPTTest("https://www.educations.com/", options);
+  pokellector = await runWPTTest("https://www.pokellector.com/", options);
```
rviscomi, line 11: Having so many external dependencies makes this test more likely to fail if any of them change. I wonder if we should have more contrived test cases on pages that we control to prevent that?

Also curious to get @tunetheweb's opinion.

max-ostapenko: I've tried to do testing previously with sites like:
- https://global-privacy-control.glitch.me/
- https://user-agent-client-hints.glitch.me/
but it doesn't cover all the cases even for privacy chapter.

That would be awesome to have some more extended library of browser API and other demos.
Then we would be able to improve testing speed and maybe run all the test-cases with each action run.

tests/privacy.test.js
max-ostapenko: It's a default behaviour with Jest (cares about suffix but not location).
And allowed to differentiate between tests and other files (but I'd rather move wpt API module to different location - `utils/wpt.js`?)

To get rid of `.test` part I could customise npm action: `"test": "jest --testMatch=**/tests/**/*.js"`
What do you think?

rviscomi: Oh I guess it's ok if it's a semantic thing with Jest.

.github/workflows/wpt-test.yml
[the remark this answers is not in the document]
rviscomi: Cool to see this working! https://github.com/HTTPArchive/custom-metrics/actions/runs/5814745249/job/15764998643

Here's one idea for how we can cleanly add this test data to the PR comments.

----

✨ WebPageTest results for privacy.js

<details>
<summary><strong>✅ Test passed for <a href="https://www.webpagetest.org/results.php?test=230809_BiDc7R_EMG">https://www.theverge.com/</a></strong></summary>

```json
{
    "privacy_wording_links": [
        {
            "keywords": [
                "POLICY"
            ],
            "text": "POLICY"
        },
        {
            "keywords": [
                "Policy"
            ],
            "text": "Policy"
        },
        {
            "keywords": [
                "PRIVACY"
            ],
            "text": "PRIVACY NOTICE"
        },
        {
            "keywords": [
                "COOKIE POLICY"
            ],
            "text": "COOKIE POLICY"
        }
    ],
    "iab_tcf_v1": {
        "present": false,
        "data": null,
        "compliant_setup": null
    },
    "iab_tcf_v2": {
        "present": false,
        "data": null,
        "compliant_setup": null
    },
    "iab_usp": {
        "present": true,
        "privacy_string": {
            "uspString": "1YNY",
            "version": 1
        }
    },
    "ads_transparency_spotlight": {
        "present": false,
        "ads_metadata": null
    },
    "document_interestCohort": false,
    "navigator_doNotTrack": true,
    "navigator_globalPrivacyControl": true,
    "document_permissionsPolicy": false,
    "document_featurePolicy": true,
    "referrerPolicy": {
        "entire_document_policy": null,
        "individual_requests": null,
        "link_relations": {
            "A": 7
        }
    },
    "media_devices": {
        "navigator_mediaDevices_enumerateDevices": false,
        "navigator_mediaDevices_getUserMedia": false,
        "navigator_mediaDevices_getDisplayMedia": false
    },
    "geolocation": {
        "navigator_geolocation_getCurrentPosition": false,
        "navigator_geolocation_watchPosition": false
    }
}
```
</details>

<details>
<summary><strong>✅ Test passed for <a href="https://www.webpagetest.org/results.php?test=230809_BiDcEC_EMK">https://www.rtl.de/</a></strong></summary>

```json
{
    "privacy_wording_links": [
        {
            "keywords": [
                "Datenschutz"
            ],
            "text": "Datenschutz"
        }
    ],
    "iab_tcf_v1": {
        "present": false,
        "data": null,
        "compliant_setup": null
    },
    "iab_tcf_v2": {
        "present": true,
        "data": {
            "cmpId": 6,
            "cmpVersion": 1,
            "gdprApplies": true,
            "tcfPolicyVersion": 2,
            "eventStatus": "cmpuishown",
            "cmpStatus": "loaded",
            "listenerId": null,
            "tcString": "CPwPGYAPwPGYAAGABCENDPCgAAAAAAAAACiQAAAAAAAA.YAAAAAAAAAAA",
            "isServiceSpecific": true,
            "useNonStandardStacks": false,
            "purposeOneTreatment": false,
            "publisherCC": "US",
            "outOfBand": {
                "allowedVendors": {},
                "disclosedVendors": {}
            },
            "purpose": {
                "consents": {},
                "legitimateInterests": {}
            },
            "vendor": {
                "consents": {},
                "legitimateInterests": {}
            },
            "specialFeatureOptins": {},
            "publisher": {
                "consents": {},
                "legitimateInterests": {},
                "customPurpose": {
                    "consents": {},
                    "legitimateInterests": {}
                },
                "restrictions": {}
            },
            "addtlConsent": "1~"
        },
        "compliant_setup": true
    },
    "iab_usp": {
        "present": false,
        "privacy_string": null
    },
    "ads_transparency_spotlight": {
        "present": false,
        "ads_metadata": null
    },
    "document_interestCohort": false,
    "navigator_doNotTrack": true,
    "navigator_globalPrivacyControl": false,
    "document_permissionsPolicy": false,
    "document_featurePolicy": true,
    "referrerPolicy": {
        "entire_document_policy": null,
        "individual_requests": null,
        "link_relations": {
            "A": 35
        }
    },
    "media_devices": {
        "navigator_mediaDevices_enumerateDevices": false,
        "navigator_mediaDevices_getUserMedia": false,
        "navigator_mediaDevices_getDisplayMedia": false
    },
    "geolocation": {
        "navigator_geolocation_getCurrentPosition": true,
        "navigator_geolocation_watchPosition": true
    }
}
```
</details>

----

So we use the `<details>` element to hide the verbose test results, and the `<summary>` element shows the pass/fail status, the name of the test page, and a link to the full test results. Let me know if you think that could be useful.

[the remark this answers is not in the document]
max-ostapenko: I've cleaned up logs formatting a bit and added logging of custom metrics to a comment (additional to logs).

Right now I don't see a way to add test results summary there without writing some custom json parser and renderer. I believe somebody should know a more straightforward solution.

max-ostapenko:
As a nice to have added [a failing test annotations](https://jestjs.io/blog/2022/04/25/jest-28#github-actions-reporter).
It's somehow obscure, but may be useful in discussions.

tests/privacy.test.js
rviscomi: @tunetheweb any thoughts?

tunetheweb:
Apologies for the radio silence from me. Been travelling a lot and not had time to review this until now.

First up, there's a lot of work here so thanks @max-ostapenko for all efforts here. 🙏

However, I'm not sure this is the right approach...

To me there are two things we could do with automated testing:
1. Test new functionality against a given website and provide a link to the WPT for that site for the custom metric file being changed. This might be temporary URL (e.g. a test site) and may not make a good long term test case, but it shows it works.
2. Have "unit tests" (or "integration" tests) to check existing custom metrics still work and a PR hasn't broken them completely.

This PR seems to address 2. but I think the original aim of #6 was more about 1.

And I do worry about the maintainability of the test cases—especially as this grows to cover each custom metric file and as sites change. Currently, if I'm reading the code correctly, this will fail if a custom metric does not have a test case, and only two of them do with this PR ads and privacy), so there's an initial effort for each person adding a small tweak to an existing custom metric if we merge this. We could pre-initialise all the custom metrics with a basic test case, but that potentially makes maintainability worse—particularly if we want meaningful test cases for each one. Or we keep them light and only add fuller coverage for people who want them to really test all their custom metrics.

So, personally, I'd prefer it if we looked at 1. more, but kept most of the work here for 2. but scaled it down somewhat.

For 1. (testing changes) I think we should have a PR template which takes a URL/URLs that is filled in by the person raising the PR, and runs the changed custom metric JavaScript file(s) against that URL/those URLs. Annoyingly, I don't think GitHub supports [fancy PR templates with fields](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#creating-issue-forms) (that's only for Issues right now AFAIK), so probably would need to regex the line out of a basic PR template and hope people fill them in right, until GitHub supports that. If no website is given we skip this test (maybe with a warning?).

For 2. (unit tests) I think we pare this back quick a bit and not depend on sites that are out of our control and may change. I would suggest we test httparchive.org (or maybe almanac.httparchive.org?) for ALL custom metrics in one go (like the actual monthly run would do for that site). We do this for every PR (except PRs that just change .md files—like README.md maybe?). This would just be a basic test to ensure all the custom metrics run, and some output is produced. I don't think we need to have checks for each custom metric as many, like ads and privacy may be empty for our sites and that's expected. Maybe we could test a couple of basic tests (e.g. those from markup.js or javascript.js) to ensure some meaningful output is produced for at least some of the custom metrics, but mainly we just want to check that a PR doesn't break everything due to a typo or bad JS.

I know this PR has been open for a while and a lot of work has gone into it, so maybe, as a first step, we could implement one of those suggested workflows in this PR? Then get it merged. Then look at the other one in it's own workflow? I'm open to doing 1 first and then 2 or vice versa.

WDYT @max-ostapenko @rviscomi ?

tunetheweb:
In not sure I understand what "1. in unit test runs skip custom metrics changed (only dist/*.js checked) but without unit tests available" means?

Does this mean "for unit test mode, don't restrict it to the files changed, but always run all the metrics, and as well as just running the custom metrics also run the any unit tests js scripts that exist against the results"?

max-ostapenko:
Clarification on todo 1:
- if dest/ads.js and dest/privacy.js was changed in PR, but no tests/ads.test.js available - we run only tests/privacy.test.js.
- if dest/cms.js was changed in PR, but no tests/cms.test.js available - we don't run any unit tests.

rviscomi:
Thanks @max-ostapenko that plan SGTM.

tunetheweb:
I was thinking more that for all PRs (except maybe simple .md edits):
- all custom metrics run on almanac.httparchive.org. Link to WebPageTest run added to PR.
- all unit tests run on  results of the above run on almanac.httparchive.org
- all custom metrics run on additional URL(s) if provided in PR issue text. Link to WebPageTest run added to PR.

If we only run the unit tests when those custom metrics are changed then it could be years between runs and they might get out of date and stop working if we change almanac.httparchive.org without thinking about this. 

Similarly changing YAML files, or dependencies might break unit tests without us realising.

Running them for every PR means it’s less like to happen.

It’s kinda like we run lots of tests on almanac.httparchive.org PRs even when they don’t touch code that should affect those tests.

max-ostapenko:
You're right, will change the unit test trigger to run on any code changes.

rviscomi:
@max-ostapenko @pmeenan BTW we should also make sure this works with our private instance. At the moment Max is using my personal WPT API key for testing 😅

max-ostapenko:
@tunetheweb @rviscomi 
So:
- We're running unit tests on every PR. WPT link to test run data in comment.
- In PR body you can define list of websites to test additionally (no unit tests run for these). The link + custom metrics values (if changed in PR) available in comment.

tests/wpt.js
```diff
@@ -0,0 +1,78 @@
+const WebPageTest = require('webpagetest');
+const fs = require('fs');
+const path = require('path');
+const { argv } = require('node:process');
+
+const is_direct_run = require.main === module;
+
+const wptServer = 'www.webpagetest.org';
```
max-ostapenko, line 8: @rviscomi @pmeenan
We need to change a server and key (value in Github secrets) to start using httparchive private instance.

max-ostapenko: @pmeenan And I would appreciate a general review of this wpt script to avoid discrepancies with the monthly crawler configuration.

package.json
```diff
@@ -19,6 +20,19 @@
   },
   "homepage": "https://github.com/HTTPArchive/custom-metrics#readme",
   "devDependencies": {
-    "fugu-api-data": "^1.15.0"
+    "fugu-api-data": "^1.15.0",
+    "jest": "^29.7.0",
+    "webpagetest": "github:max-ostapenko/webpagetest-api#post-request-support-for-runTest"
```
max-ostapenko, line 25: @rviscomi @tunetheweb could you please fork [catchpoint/WebPageTest.api-nodejs](https://github.com/catchpoint/WebPageTest.api-nodejs) under [HTTPArchive](https://github.com/HTTPArchive) so that the dependency is controlled by the org.

rviscomi: Done: https://github.com/HTTPArchive/WebPageTest.api-nodejs

tunetheweb (approved):
LGTM once we get the keys setup.

Thanks SOOO much for working on this.

pmeenan:
I have an API key allocated for this action for use with the instance at webpagetest.httparchive.org.  I currently have it assigned to the name HA_API_KEY but I can update the existing WPT_API_KEY secret if that would work better. Just need to change the action to point to the other server (or to read the server from a config).

.github/workflows/wpt-test.yml
```diff
@@ -0,0 +1,83 @@
...
+          npm install webpagetest
+          npm install jest
+
+      - name: Run WebPageTest with unit tests
+        run: |
+          echo "::group::Unit tests"
+            npm test
+          echo "::endgroup::"
+        env:
+          WPT_SERVER: "webpagetest.httparchive.org"
```
max-ostapenko, line 31: @pmeenan with new server we get an error:
```
console.error
      WPTAPIError {
        name: 'WPTAPIError',
        code: 301,
        message: 'Moved Permanently'
      }
```
Please adjust the configuration here.

.github/workflows/wpt-test.yml
```diff
@@ -0,0 +1,83 @@
...
+          WPT_API_KEY: ${{ secrets.HA_API_KEY }}
```
max-ostapenko, line 32: @pmeenan with new server we get an error:

console.error
      WPTAPIError {
        name: 'WPTAPIError',
        code: 301,
        message: 'Moved Permanently'
      }
Please adjust the configuration here.

pmeenan: Sorry, looks like the client prefers http over https and the server was set to redirect to https.  I've updated the server to allow connections over http as well.

max-ostapenko: @pmeenan Server responds with 400 now:
```
"statusCode": 400,
"statusText": "An error occurred processing your request (missing API key). If you do not have an API key you can purchase one here: https://product.webpagetest.org/api",
```
Please check the secret.

To test it:
```bash
export WPT_SERVER=webpagetest.httparchive.org
export WPT_API_KEY=INSERT_API_KEY_HERE
npm test
```

pmeenan: Sorry about the delay. Not sure why I'm not getting notifications but I'll try to check regularly until it lands.

Not 100% sure why, but it looks like if I pass the key explicitly as part of the options then it works. I submitted the change (hopefully that's ok).

max-ostapenko: @pmeenan Thanks for a hint. Looks OK indeed.

max-ostapenko:
@rviscomi no open discussions so far. Ready to merge?

rviscomi:
LGTM! Thank you so much @max-ostapenko!

rviscomi (approved):

tunetheweb:
Awesome job @max-ostapenko!

rviscomi:
Looks like this failed on a recent PR: https://github.com/HTTPArchive/custom-metrics/actions/runs/6591822240/job/17911210018?pr=96

```
Errors thrown in tests/unit-tests.test.js
    ● _ads parsing
  
      thrown: Object {
        "statusCode": 400,
        "statusText": "An error occurred processing your request (missing API key). If you do not have an API key you can purchase one here: https://product.webpagetest.org/api",
      }
```

max-ostapenko:
@rviscomi Please confirm that `HA_API_KEY` secret (the WPT token) is available.
In the debug logs [a job gets `null`](https://github.com/HTTPArchive/custom-metrics/actions/runs/6591822240/job/17958441417#step:4:7).

rviscomi:
GitHub doesn't actually show the value of the secret, but it seems like it's still there. @pmeenan could you try reapplying it?

image

image

tunetheweb:
Hmmmm this may be a flaw in our plans!

By default, secrets are deliberately not available to PRs from forks. Otherwise an attacker could just create a script to get the secret and open a PR with that script in it.

So not sure how we can get around this...

max-ostapenko:
Yeah, we need to use `pull_request_target` trigger.
Will open a PR.

One more idea:
We could fetch only particular sources from head branch.
As we read and pass `dist/*.js` sources to WPT without executing them, it may be be safe to run WPT for websites from PR description for everyone.
Let me know if this would be valuable (or point to safety issues).

BTW all first time contributors already require approval to run workflows [by default](https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks).

tunetheweb:
> We could fetch only particular sources from head branch.
> As we read and pass `dist/*.js` sources to WPT without executing them, it may be be safe to run WPT for websites from PR description for everyone.

Yeah agree that should be fine. In fact because none of the code is executed locally I think the security risk here is low so think we can checkout all the code from head and no need to limit it. In which case can do it for first run too if using `pull_request_target` I think.

The main risk is overwhelming our WebPageTest instance and/or the sites themselves with multiple tests but given we limit it to one worker (I think - @pmeenan can you confirm?) and that new users will require approval as you see there doesn't seem huge risk there either.

max-ostapenko:
PR opened #97

max-ostapenko:
[Some more workflow errors](https://github.com/HTTPArchive/custom-metrics/actions/runs/6624856217/job/17994661665?pr=96) - need to adjust PR description parsing.

tunetheweb:
Good that it ran though!

PR #93, opened 2023-08-10
Add an `iframes` custom metric

adamsilverstein:
Based on the images metric, this one enumerates iframes present on the page including their:
* url
* width & height
* naturalWidth and naturalHeight (for example when when w/h altered by CSS)
* inViewport

### DIFF
--- a/dist/iframes.js
+++ b/dist/iframes.js
@@ -1,0 +1,28 @@
+var getIframes = function (win) {
+  var iframes = [];
+  if (win) {
+    var doc = win.document;
+    var elements = doc.getElementsByTagName("iframe");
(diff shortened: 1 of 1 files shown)


tunetheweb:
Most of this already exists in the almanac custom metric:
https://github.com/HTTPArchive/custom-metrics/blob/main/metric-summary.md#iframes

We should enhance that. Though maybe there’s a good argument to move it to its own file while doing that. WDYT @rviscomi ?

tunetheweb:
Reminder @adamsilverstein . Wanna update the almanac one short term? Or take on the moving of this to markup.js at the same time as doing your updates?

PR #96, opened 2023-10-20
Detect WordPress content type

felixarntz:
In WordPress specific HTTP Archive research, we've so far typically relied on only home pages of each origin. However, home pages tend to have different layout than many other pages have, and our assumptions on e.g. performance are not segmented in any way regarding content type. It would be helpful to e.g. differentiate between specific performance patterns commonly applied on home pages vs on singular posts.

Therefore this PR introduces a WordPress specific content type detection, based on the body classes that the CMS outputs by default. The PR adds 3 fields with information:
* `template`: This is the most important field and can have one of the following values:
    * `home-blog`: The home page showing the blog (of latest posts).
    * `home-page`: The home page showing a "static front page" (as configured in WordPress).
    * `blog`: The blog (of latest posts) if configured to not also be the home page (only relevant when the home page is showing a "static front page").
    * `singular`: Any kind of singular content, e.g. a single post, page, etc.
    * `archive`: Any kind of archive content, e.g. a post type or taxonomy archive.
    * `unknown`: None of the above could be detected.
* `post_type`: The post type of the current content, or empty if not applicable based on the `template`. This field should have a value for any content type, except for a taxonomy archive.
* `taxonomy`: The taxonomy of the current content, or empty if not applicable based on the `template`. This field should have a value only for a taxonomy archive.

Some WebPageTest tests with this logic (using @westonruter's site as test candidate):
* https://www.webpagetest.org/details.php?cached=0&run=1&test=231020_BiDcVP_EVY#:~:text=E%22%2C%22securityHeadersScore%22%3A25%7D-,WPCONTENTTYPE,-%7B%22contentType%22%3A%22home%2Dblog
* https://www.webpagetest.org/details.php?cached=0&run=1&test=231020_BiDc3N_F1Y#:~:text=E%22%2C%22securityHeadersScore%22%3A25%7D-,WPCONTENTTYPE,-%7B%22contentType%22%3A%22singular%22%2C%22postType
* https://www.webpagetest.org/details.php?cached=0&run=1&test=231020_AiDcMH_EFC#:~:text=E%22%2C%22securityHeadersScore%22%3A25%7D-,WPCONTENTTYPE,-%7B%22contentType%22%3A%22singular%22%2C%22postType
* https://www.webpagetest.org/details.php?cached=0&run=1&test=231020_BiDcT5_F2K#:~:text=E%22%2C%22securityHeadersScore%22%3A25%7D-,WPCONTENTTYPE,-%7B%22contentType%22%3A%22archive%22%2C%22postType

**Test websites**:
- https://weston.ruter.net/
- https://weston.ruter.net/about/
- https://weston.ruter.net/category/wordpress/
- https://weston.ruter.net/2023/07/01/running-the-wordpress-plugin-directory-slurper/

### DIFF
--- a/dist/cms.js
+++ b/dist/cms.js
@@ -77,6 +77,11 @@ function getWordPressScripts() {
   return entries;
 }
 
+/**
+ * Detects the type of WordPress content for the current document.
+ *
+ * @returns {object} Object with fields `contentType`, `postType`, and `taxonomy`.
+ */
 function getWordPressContentType() {
   const content = {
     contentType: 'unknown',
--- a/dist/cms.js
+++ b/dist/cms.js
@@ -80,12 +80,12 @@ function getWordPressScripts() {
 /**
  * Detects the type of WordPress content for the current document.
  *
- * @returns {object} Object with fields `template`, `postType`, and `taxonomy`.
+ * @returns {object} Object with fields `template`, `post_type`, and `taxonomy`.
  */
 function getWordPressContentType() {
   const content = {
     template: 'unknown',
-    postType: '',
+    post_type: '',
     taxonomy: '',
   };
   try {
@@ -98,10 +98,10 @@ function getWordPressContentType() {
        */
       if ( bodyClass.contains( 'blog' ) ) {
         content.template = 'home-blog';
-        content.postType = 'post';
+        content.post_type = 'post';
       } else if ( bodyClass.contains( 'page' ) ) {
         content.template = 'home-page';
-        content.postType = 'page';
+        content.post_type = 'page';
       }
     } else if ( bodyClass.contains( 'blog' ) ) {
       /*
@@ -109,7 +109,7 @@ function getWordPressContentType() {
        * Only relevant if the home page contains a "static front page".
        */
       content.template = 'blog';
-      content.postType = 'post';
+      content.post_type = 'post';
     } else if ( bodyClass.contains( 'singular' ) ) {
       /*
        * Any singular content (other than the "static front page").
@@ -117,11 +117,11 @@ function getWordPressContentType() {
        */
       content.template = 'singular';
       if ( bodyClass.contains( 'page' ) ) {
-        content.postType = 'page';
+        content.post_type = 'page';
       } else if ( bodyClass.contains( 'single' ) ) {
         const postTypeClass = Array.from( bodyClass ).find( c => c.startsWith( 'single-' ) && ! c.startsWith( 'single-format-' ) );
         if ( postTypeClass ) {
-          content.postType = postTypeClass.replace( 'single-', '' );
+          content.post_type = postTypeClass.replace( 'single-', '' );
         }
       }
     } else if ( bodyClass.contains( 'archive' ) ) {
@@ -138,7 +138,7 @@ function getWordPressContentType() {
       } else if ( bodyClass.contains( 'post-type-archive' ) ) {
         const postTypeClass = Array.from( bodyClass ).find( c => c.startsWith( 'post-type-archive-' ) );
         if ( postTypeClass ) {
-          content.postType = postTypeClass.replace( 'post-type-archive-', '' );
+          content.post_type = postTypeClass.replace( 'post-type-archive-', '' );
         }
       } else {
         const taxonomyClass = Array.from( bodyClass ).find( c => c.startsWith( 'tax-' ) );
@@ -156,7 +156,7 @@ const wordpress = {
   has_embed_block: hasWordPressEmbedBlock(),
   embed_block_count: getWordPressEmbedBlockCounts(),
   scripts: getWordPressScripts(),
-  contentType: getWordPressContentType(),
+  content_type: getWordPressContentType(),
 };
 
 return {
--- a/dist/cms.js
+++ b/dist/cms.js
@@ -80,11 +80,11 @@ function getWordPressScripts() {
 /**
  * Detects the type of WordPress content for the current document.
  *
- * @returns {object} Object with fields `contentType`, `postType`, and `taxonomy`.
+ * @returns {object} Object with fields `template`, `postType`, and `taxonomy`.
  */
 function getWordPressContentType() {
   const content = {
-    contentType: 'unknown',
+    template: 'unknown',
     postType: '',
     taxonomy: '',
   };
@@ -97,10 +97,10 @@ function getWordPressContentType() {
        * or a "static front page".
        */
       if ( bodyClass.contains( 'blog' ) ) {
-        content.contentType = 'home-blog';
+        content.template = 'home-blog';
         content.postType = 'post';
       } else if ( bodyClass.contains( 'page' ) ) {
-        content.contentType = 'home-page';
+        content.template = 'home-page';
         content.postType = 'page';
       }
     } else if ( bodyClass.contains( 'blog' ) ) {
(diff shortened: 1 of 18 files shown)


felixarntz:
cc @tunetheweb @westonruter

dist/cms.js
```diff
@@ -77,11 +77,86 @@ function getWordPressScripts() {
...
 }
 
+/**
+ * Detects the type of WordPress content for the current document.
+ *
+ * @returns {object} Object with fields `contentType`, `postType`, and `taxonomy`.
+ */
+function getWordPressContentType() {
+  const content = {
+    contentType: 'unknown',
```
westonruter, line 87: Wouldn't `template` make more sense than `contentType` here?

dist/cms.js
```diff
@@ -77,11 +77,86 @@ function getWordPressScripts() {
...
+        if ( postTypeClass ) {
+          content.postType = postTypeClass.replace( 'single-', '' );
+        }
+      }
+    } else if ( bodyClass.contains( 'archive' ) ) {
+      /*
+       * Any archive (other than the blog).
+       * Either a category archive, a tag archive, a custom post type archive,
+       * or a custom taxonomy archive.
+       */
```
westonruter, line 132: What about `author` and `date` archives? Also `search`, but unlikely to be in HTTP Archive.

dist/cms.js
felixarntz: That's worth considering. I would like to not stick too much to the very granular WordPress template hierarchy, as this more about the kind of content used (e.g. "singular" vs "archive" is still at a very high level). But I like the `template` name here, as long as we're aligned this shouldn't directly need to map to a WordPress theme template identifier.

westonruter: Yeah, my concern with `contentType` is it seems more synonymous with `postType`.

dist/cms.js
felixarntz: The current approach uses the agnostic `archive` name, which would apply to any of them. Essentially, all that this does is that for post type or taxonomy archives, it also stores related metadata (post type / taxonomy).

I'm not sure there's value in storing additional data for those kinds of archives specifically, as I assume they're rarely (if ever) part of HTTP Archive data, as it's also common to discourage indexing those kinds of archives (e.g. in Yoast SEO).

felixarntz:
@tunetheweb @max-ostapenko I'm not sure what's going on with the tests in https://github.com/HTTPArchive/custom-metrics/actions/runs/6629171971/job/18007857722?pr=96
Looking at the CI failures, the test code seems to assume it needs to run some of the returned values from this check as commands (which is not the case)? 🤔 

Can you advise on that, or is there any documentation on how I should amend those tests?

tunetheweb:
Just merged a fix to `main` for this so just updated this branch. Fingers crossed it works now!

tunetheweb:
Tests now pass. And it automatically comments with the custom metrics for any URLs mentioned in the issue. Saving you the need to run these manually. You can check them to make sure your custom metric change is working as intended.

dist/cms.js
```diff
@@ -77,11 +77,86 @@ function getWordPressScripts() {
...
+  } catch ( e ) {}
+  return content;
+}
+
 const wordpress = {
   block_theme: usesBlockTheme(),
   has_embed_block: hasWordPressEmbedBlock(),
   embed_block_count: getWordPressEmbedBlockCounts(),
   scripts: getWordPressScripts(),
+  contentType: getWordPressContentType(),
```
westonruter, line 159: For consistency with the other keys:

```suggestion
  content_cype: getWordPressContentType(),
```

felixarntz: Thanks @westonruter! I've updated this to consistently use snake case (also for the inner properties) in https://github.com/HTTPArchive/custom-metrics/pull/96/commits/3cd930d0b63c68f00d2c3874012c971124ddb85b.

westonruter (approved):

felixarntz:
@tunetheweb @max-ostapenko I'm not sure I understand how the automated tests are supposed to work. Looking at the results in https://github.com/HTTPArchive/custom-metrics/pull/96#issuecomment-1777769835, for example https://webpagetest.httparchive.org/details.php?test=231024_AD_V&run=1&cached=0#:~:text=cms-,%7B%22wordpress%22%3A%7B%22block_theme%22%3Afalse%2C%22has_embed_block,-%22%3Afalse%2C%22embed_block_count%22%3A%7B%22total, I don't see the metrics from this PR being added at all.

If they were incorrect, I would assume there may be a problem with the PR, but given that the field that this PR is adding is missing entirely, I wonder whether the tests are working as intended. The `cms.wordpress` object should have a `content_type` property, but that is not the case, which makes me think this is not using the latest JS from this branch.

max-ostapenko:
@felixarntz The tested custom metric is called `_cms`, so you should look for `_cms.wordpress.content_type`.

The other `cms` custom metrics I believe are the ones used in production currently.

Screenshot 2023-10-24 at 22 41 23

tunetheweb:
They should show in "Changed custom metrics values:" section though, which I believe @max-ostapenko is looking at in #100. Will see if we can get that merged first to test on this PR.

tunetheweb:
> They should show in "Changed custom metrics values:" section though, which I believe @max-ostapenko is looking at in #100. Will see if we can get that merged first to test on this PR.

This is fixed now and seems to be working. Thanks @max-ostapenko !

dist/cms.js
```diff
@@ -77,11 +77,86 @@ function getWordPressScripts() {
...
+       * or a "static front page".
+       */
+      if ( bodyClass.contains( 'blog' ) ) {
+        content.template = 'home-blog';
+        content.post_type = 'post';
+      } else if ( bodyClass.contains( 'page' ) ) {
+        content.template = 'home-page';
+        content.post_type = 'page';
+      }
+    } else if ( bodyClass.contains( 'blog' ) ) {
```
tunetheweb, line 106: Probably a coding style preference here but there's a lot of `else if`'s in this, including nested ones, with no real common code other than the return statement.

Should we refactor to just return early to make this easier to follow?

For example:

```
    if ( bodyClass.contains( 'home' ) ) {
      /*
       * The home page, either containing the blog,
       * or a "static front page".
       */
      if ( bodyClass.contains( 'blog' ) ) {
        content.template = 'home-blog';
        content.post_type = 'post';
        return content;
      }
 
      if ( bodyClass.contains( 'page' ) ) {
        content.template = 'home-page';
        content.post_type = 'page';
        return content;
      }

      return content;
    }
    
    if ( bodyClass.contains( 'blog' ) ) {
```

Or alternatively could go for a much flatter option, but this does duplicate some code:

```
    if ( bodyClass.contains( 'home' ) && bodyClass.contains( 'blog' ) ) {
      content.template = 'home-blog';
      content.post_type = 'post';
      return content;
    }
 
    if ( bodyClass.contains( 'home' ) && bodyClass.contains( 'page' ) ) {
      content.template = 'home-page';
      content.post_type = 'page';
      return content;
    }
    
    if ( bodyClass.contains( 'blog' ) ) {
```

westonruter: IMO, the addition of many more `return` statements makes the code more complex. I like to have just a single `return` when possible.

tunetheweb: I’m definitely a fan of early return options where possible but agree this case it’s not exactly that. Though potentially it should early return if bodyClass is empty rather than waste time going through all the cases.

Thinking about the third option also potentially surfaced a flaw in the logic. If the body class contains `home` but not `blog` nor `page` do we really wanna return the default `unknown` template? Or should it try the other checks to see if it can get a better match?

felixarntz: @tunetheweb Realistically, there would never be a situation where `home` is applied without `blog` or `page`, simply because WordPress's functionality dictates that the home page is either the blog or a page. We could support returning a plain `home` value (instead of `home-blog` and `home-page`), but that wouldn't really provide value I think. So I think returning `unknown` makes more sense as this is not something that WordPress supports out of the box. If there was a `home` class only, we would know that a plugin has tampered with those classes in which case the detection isn't reliable.

tunetheweb (approved):
LGTM!

PR #97, opened 2023-10-23
Adjust tests workflow for fork PRs

max-ostapenko:
- [x] changed tests workflow to `pull_request_target` trigger

Potential vulnerabilities:
- stolen secrets.

### DIFF
--- a/.github/workflows/wpt-test.yml
+++ b/.github/workflows/wpt-test.yml
@@ -1,7 +1,7 @@
 name: Tests
 
 on:
-  pull_request:
+  pull_request_target:
     branches:
       - main
     paths-ignore:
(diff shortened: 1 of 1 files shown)


max-ostapenko:
@tunetheweb, I did minimal changes (hope I understood you correct)
And existing PRs need to rebase in order to use a new workflow.

tunetheweb:
Any ideas why it didn’t kick off a run for this PR?

.github/workflows/wpt-test.yml
```diff
@@ -1,7 +1,7 @@
 name: Tests
 
 on:
-  pull_request:
+  pull_request_target:
     branches:
```
tunetheweb, line 5: I don’t think you can use this as think `pull_request_target` is against the PR branch not against `main`. Lets delete these two lines and run it for all PR branches.

tunetheweb: Then I’m at a loss as to why this PR hasn’t triggered this test.

tunetheweb: I wonder if it has to be merged to take effect?

max-ostapenko:
I think it's because it will run once merged ([source](https://stackoverflow.com/a/77069245))
Could replicate it in [a test repo](https://github.com/max-ostapenko/test/pull/4).

tunetheweb (approved):

PR #103, opened 2023-12-06
Test LCP viewport stat

rviscomi:
Fixes for the `performance` custom metric:
- getComputedStyle already handles specificity/importance so no need to merge with inline styles
- percentage of viewport should be `* 100` to make it an int, not `/ 100`

Also outputs the LCP viewport percentage as a standalone field

**Test websites**:
- https://maxiclimber.com/
- https://fancy-confused-apparel.glitch.me/

### DIFF
--- a/dist/performance.js
+++ b/dist/performance.js
@@ -24,7 +24,7 @@ function getLcpElement() {
             resolve(naiveLcpEntry);
         }).observe({ type: "largest-contentful-paint", buffered: true });
     }).then(({ startTime, element, url, size, loadTime, renderTime }) => {
-        const cover90viewport = doesElementCoverPercentageOfViewport(element, 90);
+        const cover90viewport = doesElementCoverPercentageOfViewport(element, 0.9);
         const attributes = getAttributes(element);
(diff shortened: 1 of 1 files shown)


dist/performance.js
```diff
@@ -210,15 +190,19 @@ function getGamingMetrics(rawDoc, lcp_elem_stats) {
...
             return true;
         }
         return false;
 }
 
+function getPercentOfViewport(element) {
+        const elementBCR = element.getBoundingClientRect();
+        return (elementBCR.width * elementBCR.height * calcOcclusion(elementBCR)) / (window.innerWidth * window.innerHeight);
+}
+
```
tunetheweb: I like the fix but am not loving the changing definition of percentage here:
- percentage: 0-100
- percentOfViewport: 0 -100
- getPercentOfViewport: 0 - 1

Should we move the multiplication into `getPercentOfViewport` to at least have consistency of 0-100 everywhere?

rviscomi: I'll just make it decimals everywhere, the int stuff was just a convenience for the person who originally wrote the code

PR #104, opened 2024-01-08
Add feature flag metric

devknoll:
The [`detail` field](https://developer.mozilla.org/en-US/docs/Web/API/Performance/mark#detail) in `performance.mark` doesn't seem to be available in HTTP Archive, so this PR adds a custom metric that extracts a set of feature names according to the [User Timing specification](https://w3c.github.io/user-timing/#dfn-mark_feature_usage). This also includes support for the [previous name](https://github.com/w3c/user-timing/pull/108), since some pages are using it.

**Test websites**:

- https://angular.io/
- https://angular.dev/
- https://svelte.dev/

### DIFF
--- a/dist/aurora.js
+++ b/dist/aurora.js
@@ -57,6 +57,19 @@ function getSvelteVersion() {
   return window.__svelte?.v ? Array.from(window.__svelte?.v).toString() : null;
 }
 
+// Detects standard feature flag usage
+function getFeatureFlags() {
+  function getFeatureNames(name) {
+    return window.performance.getEntriesByName(name).map(mark => mark.detail?.feature).filter(Boolean);
+  }
+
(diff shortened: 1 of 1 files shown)


rviscomi:
FYI if you include this markdown in your PR description, the GH action will automatically test your changes on the provided URLs:

```md
**Test websites**:

- https://example.com/
```

(I've updated the PR template so it should be clearer next time)

devknoll:
Neat 👍 I edited and kicked it off so I could see :)

rviscomi:
@pmeenan since WPT has support for custom marks and measures, is this something that could be useful upstream?

rviscomi (approved):
This LGTM. The WPT question shouldn't be blocking.

PR #105, opened 2024-01-08
Set PR template to be the default

rviscomi:
New PRs aren't using the template by default because of the directory structure. In order for it to be used, the template name would need to be passed in using the `template` query parameter.

Since we want this template to be used for all new PRs, it should be moved to the top of the `.github` dir and renamed, according to the [docs](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/creating-a-pull-request-template-for-your-repository).


tunetheweb (approved):

rviscomi:
The linter error isn't applicable, merging anyway.

PR #108, opened 2024-01-18
Add custom metric to measure generated content (size and percent)

kevinfarrugia:
Adds support for `generated-content` custom metric that includes percentage and size in KB. This is helpful to recognize web pages that rely on JavaScript for generating most of the content, such as [CSR](https://www.patterns.dev/react/client-side-rendering/) apps.

As an example, see `01_test` custom metric in the following runs:
- [SSR - imkev.dev](https://www.webpagetest.org/result/240118_AiDc8Z_BZ3/1/details/)

```js
{"percent":"8.55","sizeInKB":"2.42"}	
```

- [CSR - pesuar.com/](https://www.webpagetest.org/result/240118_AiDcK3_BZ7/1/details/)

```js
{"percent":"85.40","sizeInKB":"19.92"}
```

---

**Test websites**:

- https://imkev.dev/
- https://pesuar.com/

### DIFF
--- a/dist/generated-content.js
+++ b/dist/generated-content.js
@@ -13,7 +13,7 @@ if (htmlInitial && htmlAfter) {
   generatedContent = (htmlAfterSize - htmlInitialSize) / 1024;
 }
 
-return { 
+return JSON.stringify({
   percent: generatedContentPercent.toFixed(2),
-  sizeInKB: generatedContent.toFixed(2)
-}
+  sizeInKB: generatedContent.toFixed(2),
+});
--- a/dist/generated-content.js
+++ b/dist/generated-content.js
@@ -9,11 +9,11 @@ let htmlAfter = document.documentElement.outerHTML;
 if (htmlInitial && htmlAfter) {
   let htmlInitialSize = byteSize(htmlInitial);
   let htmlAfterSize = byteSize(htmlAfter);
-  generatedContentPercent = 100 - (htmlInitialSize / htmlAfterSize) * 100;
+  generatedContentPercent = 1 - (htmlInitialSize / htmlAfterSize);
   generatedContent = (htmlAfterSize - htmlInitialSize) / 1024;
 }
 
-return JSON.stringify({
-  percent: generatedContentPercent.toFixed(2),
+return {
+  percent: generatedContentPercent.toFixed(4),
   sizeInKB: generatedContent.toFixed(2),
-});
+};
--- a/dist/generated-content.js
+++ b/dist/generated-content.js
@@ -1,0 +1,19 @@
+let generatedContentPercent = 0;
+let generatedContent = 0;
+
+const byteSize = (str) => new Blob([str]).size;
+
+let htmlInitial = $WPT_BODIES[0].response_body;
+let htmlAfter = document.documentElement.outerHTML;
+
+if (htmlInitial && htmlAfter) {
+  let htmlInitialSize = byteSize(htmlInitial);
+  let htmlAfterSize = byteSize(htmlAfter);
+  generatedContentPercent = 100 - (htmlInitialSize / htmlAfterSize) * 100;
+  generatedContent = (htmlAfterSize - htmlInitialSize) / 1024;
+}
+
+return { 
+  percent: generatedContentPercent.toFixed(2),
+  sizeInKB: generatedContent.toFixed(2)
+}


dist/generated-content.js
```diff
@@ -0,0 +1,19 @@
...
+
+const byteSize = (str) => new Blob([str]).size;
+
+let htmlInitial = $WPT_BODIES[0].response_body;
+let htmlAfter = document.documentElement.outerHTML;
+
+if (htmlInitial && htmlAfter) {
+  let htmlInitialSize = byteSize(htmlInitial);
+  let htmlAfterSize = byteSize(htmlAfter);
+  generatedContentPercent = 100 - (htmlInitialSize / htmlAfterSize) * 100;
```
rviscomi, line 12: Decimal for consistency and easier post-processing
```suggestion
  generatedContentPercent = 1 - (htmlInitialSize / htmlAfterSize);
```

dist/generated-content.js
```diff
@@ -0,0 +1,19 @@
...
+  generatedContent = (htmlAfterSize - htmlInitialSize) / 1024;
+}
+
+return JSON.stringify({
```
rviscomi, line 16: Return raw object to simplify analysis 
```suggestion
return {
```

dist/generated-content.js
```diff
@@ -0,0 +1,19 @@
...
+  percent: generatedContentPercent.toFixed(2),
+  sizeInKB: generatedContent.toFixed(2),
+});
```
rviscomi, line 19:
```suggestion
};
```

dist/generated-content.js
```diff
@@ -0,0 +1,19 @@
...
+
+if (htmlInitial && htmlAfter) {
+  let htmlInitialSize = byteSize(htmlInitial);
+  let htmlAfterSize = byteSize(htmlAfter);
+  generatedContentPercent = 100 - (htmlInitialSize / htmlAfterSize) * 100;
+  generatedContent = (htmlAfterSize - htmlInitialSize) / 1024;
+}
+
+return JSON.stringify({
+  percent: generatedContentPercent.toFixed(2),
```
rviscomi, line 17: Accommodate for change to decimal 
```suggestion
  percent: generatedContentPercent.toFixed(4),
```

rviscomi (changes requested):
Diverging a bit from the WPT metric but these changes should make it a bit easier to analyze at scale

tunetheweb:
While I agree with your suggestions @rviscomi I wonder if it's worth the divergence from the Catchpoint ones (which btw will be what will is generated from a manual run on webpagetest.httparchive.org)?

One of the reasons we're adding these is that it was a confusing that they half available in our instance (in manual runs, but not the crawl).

Maybe we should go with it as is for now and try to suggest upstream changes to theirs and, if that is accepted, then merge that change in?

pmeenan:
We are running a fork so I could just delete the catchpoint version of the
metrics. All of the server custom metrics should be disabled anyway to be
consistent with the crawl anyway.

kevinfarrugia:
Yes, it makes sense to me. There are already some small changes, so I think it's fine to not have the same implementation.

I'll just double check everything and push an update.

kevinfarrugia:
Heads up that [one of the tests](https://webpagetest.httparchive.org/result/240118_0X_E/1/details/) in the GitHub action did not run.

kevinfarrugia:
I took a look at the [test results for https://almanac.httparchive.org/en/2022/](http://webpagetest.httparchive.org/results.php?test=240118_CG_C) and it reports negative values. This is correct. The initial HTML response is larger than the one after JavaScript executes as JavaScript is used to update or remove elements, for example those containing the CSS classes `js-hide` or `js-enable`.

kevinfarrugia:
@pmeenan If I understood correctly, removing the catchpoint version of the metric should be done from the [webpagetest](https://github.com/HTTPArchive/webpagetest/tree/5aaa0101830022571cff35ad21f65049be08270f/www/settings/custom_metrics) repo.

If that is correct, is there anything else needed to include in this PR?

pmeenan:
I removed all of the server-side custom metrics from our fork (just now) so it will accurately reflect what will be measured in the crawl.

rviscomi:
@kevinfarrugia is it expected that the results for pesuar.com are `{}`?

In the top comment you noted that you got `{"percent":"85.40","sizeInKB":"19.92"}`

kevinfarrugia:
No, that is not expected. The results should be `{"percent":"85.40","sizeInKB":"19.92"}` but the [test run](https://webpagetest.httparchive.org/results.php?test=240125_WS_4) is failing. This issue occurs even on other URLs with or without the custom metric.

If I add the custom metric to webpagetest.org it works as expected.

rviscomi:
Weird ok. I can get the custom metric to succeed in a [manual test](https://webpagetest.httparchive.org/result/240126_FD_4/1/details/):

```
{"percent":"0.8502","sizeInKB":"19.32"}
```

rviscomi (approved):

rviscomi:
@pmeenan are you concerned about the test failure or do you think we can safely merge this?

pmeenan:
Likely safe to merge (worst case should be that the metric is missing).  That said, it might be worth some error handling for the case where the response_body (or even request 0) doesn't exist. I THINK that if the custom metric throws it should just be empty but that's the only part that doesn't look safe.

PR #109, opened 2024-01-31
Capture speculation rules usage

rviscomi:
Adds a `speculation_rules` field to the performance custom metric. It's an array mapping each `script[type=speculationrules]` element to its JSON contents.

Post-processing in BigQuery is required to extract meaningful stats.

---

**Test websites**:

- https://rviscomi.dev/
- https://www.scalemates.com/
- https://example.com/
- https://web.dev/


tunetheweb (approved):
Nice!

PR #113, opened 2024-04-03
Add custom metrics for the WordPress Interactivity API

adamsilverstein:
Collect data about usage of the WordPress interactivity API, which recently shipped in WordPress 6.5 (see https://make.wordpress.org/core/2024/03/04/interactivity-api-dev-note/)

Adds two new datapoints to the cms/wordpress custom metric:
* `uses_interactivity_api` - does the page uses the Interactivity API?
* `interactivity_api_region_count` - an array with the total region count as well as the region count by type.

---

**Test websites**:

- https://interactivity-api.instawp.xyz/test-post/

Current results:

### DIFF
--- a/dist/cms.js
+++ b/dist/cms.js
@@ -151,12 +151,31 @@ function getWordPressContentType() {
   return content;
 }
 
+/**
+ * Obtains data about Interactivity API elements on the page..
+ *
+ * @returns {{}[]}
+ */
+function getInteractivityAPIData() {
+  // Look for data-wp-interactive elements.
+  const interactiveElements = document.querySelectorAll('[data-wp-interactive]');
+
+  return {
+    usesInteractivityAPI: interactiveElements.length > 0,
+    interactiveElementsCount: interactiveElements.length,
+    interactiveElements: Array.from(interactiveElements).map((element) => {
+      return element.dataset.wpInteractive;
+    } )
+  };
+}
+
 const wordpress = {
   block_theme: usesBlockTheme(),
   has_embed_block: hasWordPressEmbedBlock(),
   embed_block_count: getWordPressEmbedBlockCounts(),
   scripts: getWordPressScripts(),
   content_type: getWordPressContentType(),
+  interactivity_api: getInteractivityAPIData(),
 };
 
 return {
--- a/dist/cms.js
+++ b/dist/cms.js
@@ -154,7 +154,7 @@ function getWordPressContentType() {
 /**
  * Obtains data about Interactivity API elements on the page..
  *
- * @returns {{}[]}
+ * @returns {object} Object with fields `usesInteractivityAPI`, `interactiveElementsCount`, and `interactiveElements`.
  */
 function getInteractivityAPIData() {
   // Look for data-wp-interactive elements.
--- a/dist/cms.js
+++ b/dist/cms.js
@@ -152,12 +152,12 @@ function getWordPressContentType() {
 }
 
 /**
- * Obtains data about Interactivity API elements on the page..
+ * Obtains data about Interactivity API regions on the page.
  *
  * @returns {object} Object with fields `total` and `total_by_type`.
  */
 function getInteractivityAPIRegionCounts() {
-  // Look for data-wp-interactive elements.
+  // Look for data-wp-interactive regions.
   const interactiveRegions = document.querySelectorAll('[data-wp-interactive]');
 
   // Count the regions by type.
--- a/dist/cms.js
+++ b/dist/cms.js
@@ -161,7 +161,7 @@ function getInteractivityAPIUsage() {
   const interactiveRegions = document.querySelectorAll('[data-wp-interactive]');
 
   // Count the regions by namespace.
-  const regionNamespaceCounts = {}
+  const regionNamespaceCounts = {};
   interactiveRegions.forEach( region => {
     // Extract the namespace from JSON or as a string
     let namespace = '';
--- a/dist/cms.js
+++ b/dist/cms.js
@@ -152,24 +152,24 @@ function getWordPressContentType() {
 }
 
 /**
- * Obtains data about Interactivity API regions on the page.
+ * Obtains data about Interactivity API usage on the page.
  *
- * @returns {object} Object with fields `total` and `total_by_type`.
+ * @returns {object} Object with fields `total_regions` and `total_regions_by_namespace`.
  */
-function getInteractivityAPIRegionCounts() {
+function getInteractivityAPIUsage() {
   // Look for data-wp-interactive regions.
   const interactiveRegions = document.querySelectorAll('[data-wp-interactive]');
 
-  // Count the regions by type.
-  const regionTypeCounts = [];
+  // Count the regions by namespace.
+  const regionNamespaceCounts = [];
   interactiveRegions.forEach( region => {
-    const type = region.getAttribute('data-wp-interactive');
-    regionTypeCounts[ type ] = ( regionTypeCounts[ type ] || 0 ) + 1;
+    const namespace = region.getAttribute('data-wp-interactive');
+    regionNamespaceCounts[ namespace ] = ( regionNamespaceCounts[ namespace ] || 0 ) + 1;
   })
 
   return {
-    total: interactiveRegions.length,
-    total_by_type: regionTypeCounts,
+    total_regions: interactiveRegions.length,
+    total_regions_by_namespace: regionNamespaceCounts,
   };
 }
 
@@ -189,7 +189,7 @@ const wordpress = {
   scripts: getWordPressScripts(),
   content_type: getWordPressContentType(),
   uses_interactivity_api: usesInteractivityAPI(),
-  interactivity_api_region_count: getInteractivityAPIRegionCounts(),
+  interactivity_api_usage: getInteractivityAPIUsage(),
 };
 
 return {
--- a/dist/cms.js
+++ b/dist/cms.js
@@ -174,7 +174,7 @@ function getInteractivityAPIUsage() {
     } catch ( e ) {
       namespace = regionAttribute;
     }
-    if ( '' !== namespace ) {
+    if ( 'string' === typeof namespace && '' !== namespace ) {
       regionNamespaceCounts[ namespace ] = ( regionNamespaceCounts[ namespace ] || 0 ) + 1;
     }
   })
--- a/dist/cms.js
+++ b/dist/cms.js
@@ -154,9 +154,9 @@ function getWordPressContentType() {
 /**
  * Obtains data about Interactivity API elements on the page..
  *
- * @returns {object} Object with fields `usesInteractivityAPI`, `interactiveRegionsCount`, and `interactiveRegionTypeCounts`.
+ * @returns {object} Object with fields `total` and `total_by_type`.
  */
-function getInteractivityAPIData() {
+function getInteractivityAPIRegionCounts() {
   // Look for data-wp-interactive elements.
   const interactiveRegions = document.querySelectorAll('[data-wp-interactive]');
 
@@ -168,10 +168,18 @@ function getInteractivityAPIData() {
   })
 
   return {
-    usesInteractivityAPI: interactiveRegions.length > 0,
-    interactiveRegionsCount: interactiveRegions.length,
-    interactiveRegionTypeCounts: regionTypeCounts,
+    total: interactiveRegions.length,
+    total_by_type: regionTypeCounts,
   };
+}
+
+/**
+ * Check if the page uses the interactivity API.
+ *
+ * @returns {boolean} Whether any interactivity API regions are present.
+ */
+function usesInteractivityAPI() {
+  return !!document.querySelector('[data-wp-interactive]');
 }
 
 const wordpress = {
@@ -180,7 +188,8 @@ const wordpress = {
   embed_block_count: getWordPressEmbedBlockCounts(),
   scripts: getWordPressScripts(),
   content_type: getWordPressContentType(),
-  interactivity_api: getInteractivityAPIData(),
+  uses_interactivity_api: usesInteractivityAPI(),
+  interactivity_api_region_count: getInteractivityAPIRegionCounts(),
 };
 
 return {
--- a/dist/cms.js
+++ b/dist/cms.js
@@ -163,8 +163,20 @@ function getInteractivityAPIUsage() {
   // Count the regions by namespace.
   const regionNamespaceCounts = [];
   interactiveRegions.forEach( region => {
-    const namespace = region.getAttribute('data-wp-interactive');
-    regionNamespaceCounts[ namespace ] = ( regionNamespaceCounts[ namespace ] || 0 ) + 1;
+    // Extract the namespace from JSON or as a string
+    var namespace = '';
+    const regionAttribute = region.getAttribute('data-wp-interactive');
+    try {
+      const regionData = JSON.parse( regionAttribute );
+      if ( regionData.namespace && 'string' === typeof regionData.namespace ) {
+        namespace = regionData.namespace;
+      }
+    } catch ( e ) {
+      namespace = regionAttribute;
+    }
+    if ( '' !== namespace ) {
+      regionNamespaceCounts[ namespace ] = ( regionNamespaceCounts[ namespace ] || 0 ) + 1;
+    }
   })
 
   return {
--- a/dist/cms.js
+++ b/dist/cms.js
@@ -161,7 +161,7 @@ function getInteractivityAPIUsage() {
   const interactiveRegions = document.querySelectorAll('[data-wp-interactive]');
 
   // Count the regions by namespace.
-  const regionNamespaceCounts = [];
+  const regionNamespaceCounts = {}
   interactiveRegions.forEach( region => {
     // Extract the namespace from JSON or as a string
     let namespace = '';
--- a/dist/cms.js
+++ b/dist/cms.js
@@ -164,7 +164,7 @@ function getInteractivityAPIUsage() {
   const regionNamespaceCounts = [];
   interactiveRegions.forEach( region => {
     // Extract the namespace from JSON or as a string
-    var namespace = '';
+    let namespace = '';
     const regionAttribute = region.getAttribute('data-wp-interactive');
     try {
       const regionData = JSON.parse( regionAttribute );
--- a/dist/cms.js
+++ b/dist/cms.js
@@ -154,18 +154,23 @@ function getWordPressContentType() {
 /**
  * Obtains data about Interactivity API elements on the page..
  *
- * @returns {object} Object with fields `usesInteractivityAPI`, `interactiveElementsCount`, and `interactiveElements`.
+ * @returns {object} Object with fields `usesInteractivityAPI`, `interactiveRegionsCount`, and `interactiveRegionTypeCounts`.
  */
 function getInteractivityAPIData() {
   // Look for data-wp-interactive elements.
-  const interactiveElements = document.querySelectorAll('[data-wp-interactive]');
+  const interactiveRegions = document.querySelectorAll('[data-wp-interactive]');
+
+  // Count the regions by type.
+  const regionTypeCounts = [];
+  interactiveRegions.forEach( region => {
+    const type = region.getAttribute('data-wp-interactive');
+    regionTypeCounts[ type ] = ( regionTypeCounts[ type ] || 0 ) + 1;
+  })
 
   return {
-    usesInteractivityAPI: interactiveElements.length > 0,
-    interactiveElementsCount: interactiveElements.length,
-    interactiveElements: Array.from(interactiveElements).map((element) => {
-      return element.dataset.wpInteractive;
-    } )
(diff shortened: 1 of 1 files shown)


dist/cms.js
```diff
@@ -151,12 +151,31 @@ function getWordPressContentType() {
...
 }
 
+/**
+ * Obtains data about Interactivity API elements on the page..
+ *
+ * @returns {object} Object with fields `usesInteractivityAPI`, `interactiveElementsCount`, and `interactiveElements`.
+ */
+function getInteractivityAPIData() {
+  // Look for data-wp-interactive elements.
+  const interactiveElements = document.querySelectorAll('[data-wp-interactive]');
```
felixarntz, line 161: This gets not all interactive elements, but all interactive "regions".

For interactive elements, we would need to get all elements that use any of the Interactivity API's data attributes (like `data-wp-on-...`, `data-wp-bind-...`, `data-wp-class-...`, `data-wp-init-...`, `data-wp-watch-...` etc.).

dist/cms.js
```diff
@@ -151,12 +151,31 @@ function getWordPressContentType() {
...
+
+  return {
+    usesInteractivityAPI: interactiveElements.length > 0,
+    interactiveElementsCount: interactiveElements.length,
+    interactiveElements: Array.from(interactiveElements).map((element) => {
+      return element.dataset.wpInteractive;
+    } )
```
felixarntz, line 168: This returns the data within `data-wp-interactive`, which is typically an object containing the namespace for the region, so calling it `interactiveElements` wouldn't be right.

I'm not even sure we want to provide every single interactive element (as that may end up a lot), or at least I feel like that would need to be thought through a bit more and could be something for a follow up.

I could see a value in having the interactive regions available including some of their data, such as namespaces. If we want to do that, how about calling this `interactiveRegions` and also providing some context of the element itself?

For example, instead of `return element.dataset.wpInteractive`, we could `return { tagName: element.tagName, wpInteractive: JSON.parse( element.dataset.wpInteractive ) }`? That allows us to expand what's included in the future if we come up with more useful things to include.

felixarntz (commented):
@adamsilverstein Leaving a bit of feedback. I think some of this needs to be refined a bit more in which data we want to include. Also we should differentiate between Interactivity API "regions" and _any_ elements that use the Interactivity API.

dist/cms.js
adamsilverstein: > This gets not all interactive elements, but all interactive "regions".

ok, looking at - https://developer.wordpress.org/block-editor/reference-guides/packages/packages-interactivity/packages-interactivity-api-reference/#wp-interactive I guess this is still a required wrapper so would be good to use for the boolean hasInetractivityAPI

> For interactive elements, we would need to get all elements that use any of the Interactivity API's data attributes (like data-wp-on-..., data-wp-bind-..., data-wp-class-..., data-wp-init-..., data-wp-watch-... etc.).

Ok, I will catalogue all of these are and include their data here broken out for each one, starting with counts for each type.

dist/cms.js
adamsilverstein: > For example, instead of return element.dataset.wpInteractive, we could return { tagName: element.tagName, wpInteractive: JSON.parse( element.dataset.wpInteractive ) }? That allows us to expand what's included in the future if we come up with more useful things to include.

I'm going to try to build out a demo site with all of the available elements I can find that use the Interactivity API then explore further to see what the most useful data might be.  I agree we don't need to over-collect heremaybe just counts of each namespace? 

ps. I'm curious why your code used `JSON.parse( element.dataset.wpInteractive )` - is it expected this field can contain JSON data?

adamsilverstein:
> @adamsilverstein Leaving a bit of feedback. I think some of this needs to be refined a bit more in which data we want to include. Also we should differentiate between Interactivity API "regions" and _any_ elements that use the Interactivity API.

Thanks for the feedback, I like the idea of summarizing just to the counts by type, I'm going to build out some test pages to clarify the best way to do this.

dist/cms.js
adamsilverstein: I updated the keys so it is clear this is counting regions not elements. Looking at the other attributes that might be present seems less important, I think the high level count of usage is more relevant for scaled analysis.

dist/cms.js
adamsilverstein: > This returns the data within data-wp-interactive, which is typically an object containing the namespace for the region, so calling it interactiveElements wouldn't be right.

I've renamed the keys (and updated the PR description) so they are called regions. So far the namespace is a good reference point for the type of interactive element (and hopefully plugin authors will follow the core patterns). For example on my [test post](https://interactivity-api.instawp.xyz/test-post/), the regions have the namespaces:  `core/router`, `core/query`, `core/navigation`, `core/image` and `core/search`.

adamsilverstein:
Updates:
* New naming for the returned fields to better match existing custom metrics (like embed)
* count the types (namespaces) and return that in the `interactivity_api_region_count` metric

dist/cms.js
```diff
@@ -151,12 +151,45 @@ function getWordPressContentType() {
...
+function getInteractivityAPIRegionCounts() {
+  // Look for data-wp-interactive elements.
+  const interactiveRegions = document.querySelectorAll('[data-wp-interactive]');
+
+  // Count the regions by type.
+  const regionTypeCounts = [];
+  interactiveRegions.forEach( region => {
+    const type = region.getAttribute('data-wp-interactive');
+    regionTypeCounts[ type ] = ( regionTypeCounts[ type ] || 0 ) + 1;
+  })
```
felixarntz, line 168: This grouping doesn't make sense to me. What would be the purpose of it? To my knowledge the `type` variable contains most likely a JSON string, so why would we want to group by a JSON string?

Based on https://github.com/HTTPArchive/custom-metrics/pull/113#discussion_r1550430061, I thought we were going to get the counts per type of Interactivity API attribute - though we could also do that in a separate PR.

I'm just questioning why we would want to get counts grouped by this attribute value?

dist/cms.js
```diff
@@ -151,12 +151,45 @@ function getWordPressContentType() {
...
+}
+
 const wordpress = {
   block_theme: usesBlockTheme(),
   has_embed_block: hasWordPressEmbedBlock(),
   embed_block_count: getWordPressEmbedBlockCounts(),
   scripts: getWordPressScripts(),
   content_type: getWordPressContentType(),
+  uses_interactivity_api: usesInteractivityAPI(),
+  interactivity_api_region_count: getInteractivityAPIRegionCounts(),
```
felixarntz, line 192: I think I prefer your original approach of putting an object in here that is extendable with more data in the future, to avoid potentially having lots of Interactivity API top-level properties in here in the future.

I think `uses_interactivity_api` in here on its own makes sense, but for anything more granular about the Interactivity API maybe we can use an object like `interactivity_api_usage`? For now it could have things like `regionCount` etc., and the list could be expanded as needed in the future.

dist/cms.js
adamsilverstein: > Since the technical concept is called "namespace" by the API, we should probably name this here accordingly, something like regionCountByNamespace?

Yea, good point - I was having fun calling it a type, but it really is a namespace. Will update naming to use namespace instead of type.

dist/cms.js
adamsilverstein: > maybe we can use an object like interactivity_api_usage? For now it could have things like regionCount etc., and the list could be expanded as needed in the future

Yes, good suggestion, I will update the naming here.

adamsilverstein: updated to `interactivity_api_usage `, also renamed return variable names a bit

dist/cms.js
adamsilverstein: switched everything to `namespace`. hopefully plugins follow the core naming pattern for namespace even though it isn't required and we get nice strings that let us count usage and presence on the page, the current approach works well for core blocks anyway.

felixarntz: I think we need to account for both formats, as that's how the API is defined. We can't assume the `data-wp-interactive` attribute always contains a namespace string, otherwise we get incorrect data. So I think this code will need to check if the attribute contains a JSON string that resolves to an object with `namespace` property, or simply a single (non-JSON) string, which would need to be interpreted as the namespace.

adamsilverstein: makes sense, wish we had a live example of this use; in the meantime I will create a mocked version to test.

pmeenan:
FYI, the April crawl kicks off Wednesday. Not sure if you want this metric in by then but, if so, it'll need to land today so I can update the agents before it kicks off.

dist/cms.js
```diff
@@ -151,12 +151,45 @@ function getWordPressContentType() {
...
+ * @returns {object} Object with fields `total_regions` and `total_regions_by_namespace`.
+ */
+function getInteractivityAPIUsage() {
+  // Look for data-wp-interactive regions.
+  const interactiveRegions = document.querySelectorAll('[data-wp-interactive]');
+
+  // Count the regions by namespace.
+  const regionNamespaceCounts = [];
+  interactiveRegions.forEach( region => {
+    const namespace = region.getAttribute('data-wp-interactive');
```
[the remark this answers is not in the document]
adamsilverstein: good point, I'm going to use a slightly different code approach with the same effect.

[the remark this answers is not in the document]
adamsilverstein: Added in 16fc65e2c252c718da275cabadb295f86889430e

dist/cms.js
adamsilverstein: I added a mocked up test region that uses the JSON format on the test page: https://interactivity-api.instawp.xyz/test-post/ - tested and new namespace as part of JSON was detected properly

dist/cms.js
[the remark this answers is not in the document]
adamsilverstein: @westonruter - I added a mocked up test region that uses the JSON format on the test page: https://interactivity-api.instawp.xyz/test-post/ 

image

Tested running the code in the console to validate it properly captures the namespace from the JSON.

tunetheweb:
@adamsilverstein I updated the first comment to the correct heading needed to trigger the test ("Test websites:" rather than "Test page"). So if you push a commit now it should kick that off for your test site.

dist/cms.js
```diff
@@ -151,12 +151,57 @@ function getWordPressContentType() {
...
+ */
+function getInteractivityAPIUsage() {
+  // Look for data-wp-interactive regions.
+  const interactiveRegions = document.querySelectorAll('[data-wp-interactive]');
+
+  // Count the regions by namespace.
+  const regionNamespaceCounts = [];
+  interactiveRegions.forEach( region => {
+    // Extract the namespace from JSON or as a string
+    var namespace = '';
```
westonruter, line 167: Nit:

```suggestion
    let namespace = '';
```

adamsilverstein: right, thanks!

adamsilverstein:
> @adamsilverstein I updated the first comment to the correct heading needed to trigger the test ("Test websites:" rather than "Test page"). So if you push a commit now it should kick that off for your test site.

@tunetheweb - Ah, thanks! I probably changed that from the template because I was only linking to one website (and I didn't realize this triggered the test).

I noticed when the test action ran, the results don't include the full region counts, ie:
image
`total_regions_by_namespace` shows as empty.

Is that because the output doesn't show the full data, or something off with the code? The code did work correctly when I ran in the console.  I'll try running in WPT through the UI.

tunetheweb:
> Is that because the output doesn't show the full data, or something off with the code? The code did work correctly when I ran in the console. I'll try running in WPT through the UI.

It's because we can't `JSON.stringify` an Array

The same thing happens in DevTools if you do this:

```js
function getInteractivityAPIUsage() {
  // Look for data-wp-interactive regions.
  const interactiveRegions = document.querySelectorAll('[data-wp-interactive]');

  // Count the regions by namespace.
  const regionNamespaceCounts = []
  interactiveRegions.forEach( region => {
    // Extract the namespace from JSON or as a string
    let namespace = '';
    const regionAttribute = region.getAttribute('data-wp-interactive');
    try {
      const regionData = JSON.parse( regionAttribute );
      if ( regionData.namespace && 'string' === typeof regionData.namespace ) {
        namespace = regionData.namespace;
      }
    } catch ( e ) {
      namespace = regionAttribute;
    }
    if ( '' !== namespace ) {
      regionNamespaceCounts[ namespace ] = ( regionNamespaceCounts[ namespace ] || 0 ) + 1;
    }
  })

  console.log(regionNamespaceCounts, JSON.stringify(regionNamespaceCounts));

  return JSON.stringify({
    total_regions: interactiveRegions.length,
    total_regions_by_namespace: regionNamespaceCounts,
  });
}
getInteractivityAPIUsage();
```

Will suggest something.

dist/cms.js
```diff
@@ -151,12 +151,57 @@ function getWordPressContentType() {
...
+ * Obtains data about Interactivity API usage on the page.
+ *
+ * @returns {object} Object with fields `total_regions` and `total_regions_by_namespace`.
+ */
+function getInteractivityAPIUsage() {
+  // Look for data-wp-interactive regions.
+  const interactiveRegions = document.querySelectorAll('[data-wp-interactive]');
+
+  // Count the regions by namespace.
+  const regionNamespaceCounts = [];
```
tunetheweb, line 164: Should this be an object? So you get names for each subarray? Which will be useful when stringifying it to save in BigQuery:

```suggestion
  const regionNamespaceCounts = {};
```

That, coupled with the change below to `JSON.Stringify` leads to this on your sample page:

```'{"total_regions":6,"total_regions_by_namespace":{"core/navigation":6}}'```

File: dist/cms.js [hunk 8]
```diff
@@ -151,12 +151,57 @@ function getWordPressContentType() {
...
+    }
+    if ( '' !== namespace ) {
+      regionNamespaceCounts[ namespace ] = ( regionNamespaceCounts[ namespace ] || 0 ) + 1;
+    }
+  })
+
+  return {
+    total_regions: interactiveRegions.length,
+    total_regions_by_namespace: regionNamespaceCounts,
+  };
```
tunetheweb, line 185:
```suggestion
  return JSON.stringify({
    total_regions: interactiveRegions.length,
    total_regions_by_namespace: regionNamespaceCounts,
  });
```

tunetheweb (changes requested):

dist/cms.js
felixarntz: @tunetheweb Why is this necessary? I personally would find it odd to find stringified JSON within data that's already JSON. That's always been a bit mysterious to me with HTTP Archive data (also in `payload` column), why do we have to JSON-decode twice?

dist/cms.js
```diff
@@ -151,12 +151,57 @@ function getWordPressContentType() {
...
+    const regionAttribute = region.getAttribute('data-wp-interactive');
+    try {
+      const regionData = JSON.parse( regionAttribute );
+      if ( regionData.namespace && 'string' === typeof regionData.namespace ) {
+        namespace = regionData.namespace;
+      }
+    } catch ( e ) {
+      namespace = regionAttribute;
+    }
+    if ( '' !== namespace ) {
```
felixarntz, line 177: Note that it's also possible for `data-wp-interactive` to contain nothing, in that case it would just be a boolean attribute. So to avoid e.g. something like `true` in the data, we should include something like:
```suggestion
    if ( 'string' === typeof namespace && '' !== namespace ) {
```

felixarntz (commented):
Thanks @adamsilverstein, this looks solid to me. Just one small point of follow up feedback.

felixarntz (approved):
Thanks @adamsilverstein!

dist/cms.js
adamsilverstein: added

dist/cms.js
adamsilverstein: Same question. The github-action did return the expected data:

image

dist/cms.js
adamsilverstein: I also added some test blocks to the test page and verified the counts still look right:
image

adamsilverstein:
> It's because we can't JSON.stringify an Array

@tunetheweb Ah, right. This has caused me problems before (and yet, I tried again). 

> `return JSON.stringify(`

I'm not clear on this bit: we don't stringify other returns, for example the embed block data returned from `getWordPressEmbedBlockCounts` above... why is JSON.stringify needed here? (the data structure seems very similar to the embed data)

dist/cms.js
felixarntz: +1 to what @adamsilverstein shared. There isn't actually an array in the data, it's only objects.

But even generally, I'm not sure what's the limitation of using `JSON.stringify()` on an array? 🤔

tunetheweb:
Ah sorry you don't need to `JSON.stringify` each function separately. I was testing with just the `getInteractivityAPIUsage` function call, and not the whole thing.

You SHOULD however `JSON.stringify` at the end.

```js
return JSON.stringify({
  wordpress
};
```

JavaScript objects are NOT the same as JSON. They look very similar, but not quick the same. A JavaScript object is a structured object which can include references. JSON is just a string in a particular format. Not even JavaScript object can be cleanly converted to a JSON string.

Which is what we saw earlier, and what we can repeat here with this smaller bit of code:

```js
const myArray = [];
myArray['subArray1'] = [1,2,3];
myArray['subArray2'] = [4,5,6];
console.log(myArray);
console.log(JSON.stringify(myArray));
```

image

The first line prints named arrays, the second just an empty array since you can't convert named arrays to JSON (they are basically parameters on the array variable, rather than values in the array).

The length of that array is 0 btw:

image

Because you are NOT pushing values in that array, you are setting attributes on that array. And when you stringify an array, it only looks at the values, not the attributes.

This however does work:

```js
const myArray = [];
myArray[0] = [1,2,3];
myArray[1] = [4,5,6];
console.log(myArray);
console.log(JSON.stringify(myArray));
```

image

Anyway, HTTP Archive stores these as Strings. In most cases it will convert these automatically which is why you don't NEED it here. However I advise to use `JSON.stringify()` at the end. Why? Because then when you test this in the console you'll also see when it's missing stuff you expected to in there rather than be confused why WebPageTest's magic string conversion (which isn't magic at all) did something different than you expected.

dist/cms.js
tunetheweb: >Why is this necessary? I personally would find it odd to find stringified JSON within data that's already JSON. 

Yes as mentioned below you only need to stringify the outermost return. Apologies for the confusion.

> That's always been a bit mysterious to me with HTTP Archive data (also in payload column), why do we have to JSON-decode twice?

Ah that's legacy thing. Some of the older custom metrics are double escaped. Yeah I don't like that either. We shouldn't do that anymore for newer custom metrics. Dunno why we did to be honest - before my time.

> +1 to what @adamsilverstein shared. There isn't actually an array in the data, it's only objects.

It was (incorrectly) originally set to be an Array as mentioned in my comment above.

> But even generally, I'm not sure what's the limitation of using JSON.stringify() on an array? 🤔

As described below the "limitation" is when you set up an array, then using "named" array entries, which aren't actually array entries at all:

```js
const myArray = []
myArray[1] = 'value'; // this is fine
myArray['key'] = 'value'
```

That last line is incorrect and actually is shorthand for `myArray.key = 'value';'
You've basically set an attribute on the array and not added a value to the array.
When you convert the array to a string, it doesn't include attributes (it would if it was an object, but it's not, it's an array). It only exports entries to the string and in this example `myArray` only has one entry not two (as can be seen with `myArray.length`).

Why does JavaScript allow you to set attributes on an Array when it's almost certainly done in error? 🤷‍♂️ Ultimately everything in JavaScript is an object so you can always set attributes.

Why does JavaScript have a similar syntax for setting array values and array attributes? 🤷‍♂️🤷‍♂️ Just one of the many fun things with JavaScript!

tunetheweb (approved):
As discussed I think we should JSON.stringify the final value so when testing this in the console you can more clearly see when it's not including stuff you thought it would. That way it will seem less like HTTP Archive magic and more like JavaScript weirdness 😁

However, lets merge this for now and worry about that after this run.

dist/cms.js
```diff
@@ -151,12 +151,57 @@ function getWordPressContentType() {
...
+    let namespace = '';
+    const regionAttribute = region.getAttribute('data-wp-interactive');
+    try {
+      const regionData = JSON.parse( regionAttribute );
+      if ( regionData.namespace && 'string' === typeof regionData.namespace ) {
+        namespace = regionData.namespace;
+      }
+    } catch ( e ) {
+      namespace = regionAttribute;
+    }
```
tunetheweb, line 176: @pmeenan this seems to error a lot of JSON.parse as quite often it's just a plain string rather than JSON.

Any concerns with this? Like should we test it begins with `{` before attempting to parse to reduce the number of exceptions? Or is the cost of those similar anyway?

pmeenan: I can't imagine the cost of a single exception handler, even in a loop that would get run a bunch of times, would be anything more than a rounding error. No concerns here.

pmeenan: This seems more legible than trying to detect conditions and maybe throw anyway so my vote would be to leave it as-is.

tunetheweb: Cool. Will likely be on very small strings anyway. Will merge.

PR #117, opened 2024-05-03
Privacy 2024 - CNAMEs and cleanup

max-ostapenko:
Removed:
- [x] FLoC, long closed origin trial
- [x] Ads Transparency Spotlight schema, deprecated

Updated:
- [x] privacy policy links tracking (languages extended, matching performance)
- [x] docs with a mention of WPT data objects

Added:
- [x] list of request hostnames with CNAME records.

---

**Test websites**:

- https://www.adobe.com
- https://www.amazon.com
- https://www.google.com
- https://dziennikzachodni.pl

### DIFF
--- a/dist/privacy.js
+++ b/dist/privacy.js
@@ -130,7 +130,7 @@ return JSON.stringify({
     }
 
     const privacy_links = Array.from(document.querySelectorAll('a')).filter(a =>
-     pattern.test(a.innerText) // && pattern.test(a.href) && (isSameDomain(a.href) || isRelativeURL(a.href))
+      pattern.test(a.innerText) // && pattern.test(a.href) && (isSameDomain(a.href) || isRelativeURL(a.href))
     ).map(
       a => ({
         //href: a.href
--- a/dist/privacy.js
+++ b/dist/privacy.js
@@ -107,34 +107,10 @@ return JSON.stringify({
     }
     const pattern = new RegExp(`(?:${keywords})`, 'gi');
 
-    const extractDomainFromHostname = (hostname) => {
-      const domainRegex = /^(?:.*\.)?(.+\..+)$/;
-      const matches = hostname.match(domainRegex);
-      return matches ? matches[1] : null;
-    }
-
-    const extractDomainFromURL = (url) => {
-      const domainRegex = /^(?:https?:)?\/\/(?:.*\.)?([^:\/&?]+\.[^\/:&?]+)/;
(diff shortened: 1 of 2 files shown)


dist/privacy.js
```diff
@@ -346,5 +346,26 @@ return JSON.stringify({
...
+  request_hostnames_with_cname: (() => {
+    let results = {};
+
+    for (const request of $WPT_REQUESTS) {
+      request_hostname = (new URL(request.url)).hostname;
+
+      for (const [origin, dns_info] of Object.entries($WPT_DNS)) {
+        dns_hostname = (new URL(origin)).hostname;
+
+        if (dns_hostname === request_hostname) {
```
pmeenan, line 363: Maybe also check to make sure the cname is not the same as the hostname (not sure how/why that should happen but the test results made it look like it is).

dist/privacy.js
```diff
@@ -346,5 +346,26 @@ return JSON.stringify({
...
+          results[dns_hostname] = dns_info.results.canonical_names[0];
```
pmeenan, line 364: Do you want all of the CNAMEs in the chain or just the first?

pmeenan: If 0 is the end of the chain then it's probably fine. Just wanted to make sure you weren't losing data if it was hidden behind another cname.

tunetheweb (approved):
LGTM. This good to merge @max-ostapenko ?

max-ostapenko:
All is completed.
@tunetheweb please merge

dist/privacy.js
```diff
@@ -291,10 +315,31 @@ return JSON.stringify({
...
+   * List of hostnames with CNAME record
+   */
+  request_hostnames_with_cname: (() => {
+    let results = {};
+
+    for (const request of $WPT_REQUESTS) {
+      request_hostname = (new URL(request.url)).hostname;
+
+      for (const [origin, dns_info] of Object.entries($WPT_DNS)) {
+        dns_hostname = (new URL(origin)).hostname;
```
bstandaert-wustl, line 335: @max-ostapenko I'm obviously late to reviewing this, but `new URL` can throw on an invalid URL, which would cause the entire metric recording to fail since there's no error handling. This should probably be wrapped in a try/catch.

(I have no examples of this actually occurring)

max-ostapenko: Thanks

tunetheweb: Oh btw I fixed this in #130

PR #119, opened 2024-05-19
Privacy 2024 - Implement CCPA link metric

bstandaert-wustl:
---

**Test websites**:

- https://ebay.com
- https://walmart.com


max-ostapenko:
@bstandaert-wustl please move it into privacy chapter custom metric under `dist/privacy.js` file

max-ostapenko:
Number of the websites that make sense to test for CCPA compliance is quite small compared to the dataset, particularly considering the crawlers run from Virginia, not California.

Could we define the eligible set of sites to optimize it, e.g. using locale, TLD?

pmeenan:
FWIW, The crawl runs from most of the US regions of Google Cloud so there's SOME California testing (~1/7 of the crawl) but it's not deterministic and hard to say how IP geolocation works for any of the regions.

bstandaert-wustl:
@max-ostapenko From https://petsymposium.org/popets/2022/popets-2022-0030.pdf:

> the CCPA applies to all websites doing business in California, regardless of the domicile of their business or the language of their website.

I don't think we have a reasonable way to filter on "doing business in California". Perhaps we could rank sites by monthly traffic and take only the top fraction (since the CCPA has a minimum user threshold).

bstandaert-wustl:
Per the discussion in Slack, @umariqbal recommends testing this on all sites and framing it as “prevalence of enforcement mechanisms” rather than compliance: https://httparchive.slack.com/archives/C023K97SR8U/p1716560204137519

Are there any other changes you recommend making to this before the deadline?

dist/ccpa-link.js
```diff
@@ -0,0 +1,67 @@
...
+]
+
+// https://petsymposium.org/popets/2022/popets-2022-0030.pdf page 627
+const CCPAExclusionPhrases = [
+    'terms',
+    'user agreement',
+    'service agreement',
+    'conditions of use',
+    'terms of usage',
+    'privacy notice',
```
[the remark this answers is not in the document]
max-ostapenko: @bstandaert-wustl sorry, was unsubmitted. Can you see it now?

dist/ccpa-link.js
```diff
@@ -0,0 +1,67 @@
...
+    'login',
+    'terms and conditions apply',
+    'accessibility',
+    'your data in search',
+    'shield',
+    'promo',
+    'campaign',
+    'deal',
+    'ad choice',
+    'january',
```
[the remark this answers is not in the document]
bstandaert-wustl: The list comes from this paper: https://petsymposium.org/popets/2022/popets-2022-0030.pdf. They ran a test crawl and then manually checked the results, so all of the exclusion phrases caused a false positive on at least one site. I don't think they identify which sites though.

dist/ccpa-link.js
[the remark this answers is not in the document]
bstandaert-wustl: I can, thanks. I don't know where this scenario occurs, but the authors of the linked paper claim it does.

tunetheweb:
@max-ostapenko @bstandaert-wustl what's the latest on this? We need to merge this today if we want to include it in the Web Almanac 2024.

bstandaert-wustl:
@tunetheweb I think the remaining questions are minor - if we aren't able to resolve them before your deadline, can you go ahead and merge this?

max-ostapenko:
Yeah, I see no blockers. 👍

tunetheweb (approved):

tunetheweb:
Got merge conflicts now @max-ostapenko . Can you resolve?

dist/privacy.js
```diff
@@ -341,5 +341,72 @@ return JSON.stringify({
...
+      'terms of sale',
+      'login',
+      'terms and conditions apply',
+      'accessibility',
+      'your data in search',
+      'shield',
+      'promo',
+      'campaign',
+      'deal',
+      'ad choice',
```
max-ostapenko, line 382:
```suggestion
```

@bstandaert-wustl Isn't `Your ad choices` a valid CCPA link?

dist/privacy.js
```diff
@@ -321,7 +321,7 @@ return JSON.stringify({
       'geolocation.+watchPosition'
     ),
   },
-
+
```
max-ostapenko, line 324:
```suggestion

```
@bstandaert-wustl linting edit

max-ostapenko (changes requested):

dist/privacy.js
max-ostapenko: Seems it's some alternative framework.

max-ostapenko:
@tunetheweb please merge

max-ostapenko (approved):

dist/privacy.js
bstandaert-wustl: I believe it's usually for https://en.wikipedia.org/wiki/AdChoices

PR #120, opened 2024-05-23
[Performance 2024] Added bytesPerPixel (BPP) property to Images custom metric

kevinfarrugia:
Add `bytesPerPixel` custom metric. This is calculated as (filesize in bits) / (image display height * image display width)

**[Note that it would return `0` for cross-origin images that do not include the `timing-allow-origin` header.](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceResourceTiming/encodedBodySize#value)**

---
**Test websites**:

- https://web.dev
- https://almanac.httparchive.org/

### DIFF
--- a/dist/Images.js
+++ b/dist/Images.js
@@ -24,6 +24,11 @@ var wptImages = function (win) {
               el.getBoundingClientRect().right >= 0 &&
               el.getBoundingClientRect().top <= window.innerHeight &&
               el.getBoundingClientRect().left <= window.innerWidth,
+            bytesPerPixel:
+              (performance.getEntriesByName(el.currentSrc)[0]
+                ?.encodedBodySize *
+                8) /
+              (el.width * el.height),
           });
         }
       }


tunetheweb:
A few questions:
1. Why encoded body size instead of decoded body size? If specifically interested in encoded I think we should include that in the name.
2. Does this belong in performance or images custom metrics?
3. I noticed this is already being calculated (with decoded body size) in [responsive_images.js](https://github.com/HTTPArchive/custom-metrics/blob/main/dist/responsive_images.js) custom metric. Could that be used? Or extended if you really want encoded body size.

kevinfarrugia:
1. Fair point. It isn't clear on https://github.com/w3c/largest-contentful-paint/issues/86 whether it is calculated on encoded or decoded body size. Other tools (for example https://www.phpied.com/quick-bpp-image-entropy-check/) seem to use `decodedBodySize` so we followed suit.
2. I'm not sure of the standard being followed TBH. Originally we were going to place it in its own file but thought in the Images.js mde the most sense. The performance.js file doesn't include all images, only LCP elements—TBF we are after BPP on LCP image.
3. Will check. At a glance it seems to be correct and also checks `wptRequest.response_headers` for cross-origin resources.

tunetheweb:
Oh I  missed that this was being added to images.js, I presumed it was for performance.js. Oops. OK seems fine in images.js (unless easy to add to responsive_images.js). Ideally we'd merge that and responsive-images.js I guess, but that seems like a bigger change we'd want to do with the Media chapter.

tunetheweb:
> Fair point. It isn't clear on https://github.com/w3c/largest-contentful-paint/issues/86 whether it is calculated on encoded or decoded body size.

I would imagine it's decoded size.

> Other tools (for example https://www.phpied.com/quick-bpp-image-entropy-check/) seem to use decodedBodySize so we followed suit.

But you didn't. You used _encoded_ size?

kevinfarrugia:
> > Other tools (for example https://www.phpied.com/quick-bpp-image-entropy-check/) seem to use decodedBodySize so we followed suit.
> 
> But you didn't. You used _encoded_ size?

Sorry I mixed them up. The other tools use `encodedBodySize` but TBF I am not sure which is correct.

PR #123, opened 2024-05-28
Added a custom metric to store local and session storage keys and val…

shaoormunir:
This PR adds a new custom metric to capture and return all key-value pairs from `localStorage` and `sessionStorage` of the first-party website.

## Changes

- Added `getLocalStorageKeyValues` function to retrieve all key-value pairs from localStorage.
- Added `getSessionStorageKeyValues` function to retrieve all key-value pairs from sessionStorage.
- Aggregated the results of both functions into a single object `storageData` to return for storage.

Testing was done on the `webpagetest.org.`

---

**Test websites**:

- https://tmz.com
- https://cnn.com

### DIFF
--- a/dist/local_session_storage_key_values.js
+++ b/dist/local_session_storage_key_values.js
@@ -1,9 +1,19 @@
 function getLocalStorageKeyValues() {
-  return { ...localStorage };
+  return Object.keys(localStorage)
+    .reduce((obj, key) => {
+      const value = localStorage.getItem(key);
+      obj[key] = value ? value.substring(0, 256) : value;
+      return obj;
+    }, {});
 }
 
 function getSessionStorageKeyValues() {
-  return { ...sessionStorage };
+  return Object.keys(sessionStorage)
+    .reduce((obj, key) => {
+      const value = sessionStorage.getItem(key);
+      obj[key] = value ? value.substring(0, 256) : value;
+      return obj;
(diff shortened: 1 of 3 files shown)


tunetheweb:
This has the potential to be quite large (10MB or more per domain). I think we should limit each value to the first X bytes (for some value of X) and/or exclude large base64 pieces of data (which could be encoded pictures).

WDYT?

shaoormunir:
@tunetheweb I agree with the size issue, will add a few size and length constraints while storing values.

tunetheweb:
Any update @shaoormunir  as we have one week to get this reviewed and merged and have a growing queue of other metrics too.

shaoormunir:
@tunetheweb updated to only store the first 256 bytes for each local and session storage item

tunetheweb:
As discussed in https://github.com/HTTPArchive/custom-metrics/pull/124#issuecomment-2155352026 we should remove the number of calls.

I'm also questioning the value of the local and session storage data. Even with the 256 cap, it still can be quite big looking at the CNN example for example. That makes every single call for custom metrics, or payloads bigger, slower, and more expensive for every single user of the HTTP Archive. Do we have concrete plans for this data to justify that?

rviscomi:
Closing due to inactivity but feel free to reopen if you're able to address the feedback

PR #125, opened 2024-06-01
Security 2024 Updated security.txt Metric

JannisBush:
Updated custom metric for https://github.com/HTTPArchive/almanac.httparchive.org/issues/3604

Description of the changes:
Update the parsing of `.well-known/security.txt` to take all new defined fields into account, save undefined/future/custom fields and a basic parsing of whether the file is valid (required fields exist and no field that is only allowed to occur once occurs more than once).

---

**Test websites**:

- https://example.com/
- https://securitytxt.org/
- https://facebook.com/
- https://slack.com

### DIFF
--- a/dist/well-known.js
+++ b/dist/well-known.js
@@ -169,9 +169,11 @@ return Promise.all([
         } else if (line.startsWith('CSAF: ')) {
           data['csaf'].push(line.substring(6).trim());
         } else {
-          let [name, value] = line.split(': ');
-          if (name && value) {
-            data['other'].push([name.trim(), value.trim()]);
+          if (!line.startsWith('#')) {
+            let [name, value] = line.split(': ');
+            if (name && value) {
+              data['other'].push([name.trim(), value.trim()]);
+            }
           }
         }
       }   
--- a/dist/well-known.js
+++ b/dist/well-known.js
@@ -176,8 +176,18 @@ return Promise.all([
             }
           }
         }
-      }   
-      // Optional
+      }
+      // Required fields exist
+      data['all_required_exist'] = (data['contact'].length && data['expires'].length) > 0;
+      // Fields that are only allowed once do not occur twice
+      data['only_one_requirement_broken'] = false;
+      for (let field of ['expires', 'preferred_languages']) {
+        if (data[field].length > 1) {
+          data['only_one_requirement_broken'] = true;
+        }
+      }
+      // Valid: Required fields exist and only one requirement is not broken. Does not check value content at the moment (e.g., if expires is a valid ISO 8601 date).
+      data['valid'] = data['all_required_exist'] && (!data['only_one_requirement_broken'])
       return data;
     });
   }),
--- a/dist/well-known.js
+++ b/dist/well-known.js
@@ -139,17 +139,43 @@ return Promise.all([
       if (text.startsWith('-----BEGIN PGP SIGNED MESSAGE-----')) {
         data['signed'] = true;
       }
-      for(let line of text.split('\n')) {
-        if (line.startsWith('Canonical: ')) {
-          data['canonical'] = line.substring(11);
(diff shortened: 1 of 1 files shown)


dist/well-known.js
```diff
@@ -139,17 +139,55 @@ return Promise.all([
...
+      data['policy'] = [];
+      data['hiring'] = [];
+      data['csaf'] = [];
+      data['other'] = []; // [(name, value)]
+      for (let line of text.split('\n')) {
+        if (line.startsWith('Contact: ')) {
+          data['contact'].push(line.substring(9).trim());
         } else if (line.startsWith('Expires: ')) {
-          data['expires'] = line.substring(9);
+          data['expires'].push(line.substring(9).trim());
```
SaptakS, line 156: Why do these need to be arrays instead of just string value?

JannisBush: The specification states that some fields (expires and preferred-languages) are only allowed to occur once, other fields are allowed to occur several times.

By using arrays, we can easily count how often fields occur multiple times or not and for a basic validity check validate that the fields that are only allowed to occur once do not occur multiple times.
Also we are not simply choosing the first or the last one if multiple entries exist. We could also do a string concatenation in such cases, however, that would make it more difficult to split the values once again if we want to ask whether some fields occur more than once.

tunetheweb (changes requested):
This feels quite verbose when so many fields are empty:

```json
        "/.well-known/security.txt": {
            "found": false,
            "data": {
                "status": 500,
                "redirected": false,
                "url": "https://example.com/.well-known/security.txt",
                "signed": false,
                "contact": [],
                "expires": [],
                "encryption": [],
                "acknowledgments": [],
                "preferred_languages": [],
                "canonical": [],
                "policy": [],
                "hiring": [],
                "csaf": [],
                "other": [],
                "all_required_exist": false,
                "only_one_requirement_broken": false,
                "valid": false
            }
        },
```

Can we only include the fields when they are present to reduce the storage and query size?

JannisBush:
> This feels quite verbose when so many fields are empty:

> Can we only include the fields when they are present to reduce the storage and query size?

Don't know about the storage and query size of empty arrays. The keys are always the same so there might be some optimization possible.

However, I adapted the query to only keep non-empty fields. Hope that does not make the query more complex.

JannisBush:
```json
        "/.well-known/security.txt": {
            "found": false,
            "data": {
                "status": 404,
                "redirected": false,
                "url": "https://example.com/.well-known/security.txt",
                "signed": false,
                "other": [
                    [
                        "background-color",
                        "#f0f0f2;"
                    ],
                    [
                        "margin",
                        "0;"
                    ],
                    [
                        "padding",
                        "0;"
                    ],
```

Is not great, as the inline CSS is detected as Other directives.
I now also save the content-type (MUST be text/plain according to the spec but unclear if all sites follow the spec, there are probably quite some sites that do not set any content-type header 🤔)
Additionally, I only save the data is the status is of type `ok` (`r.ok` has to be true).
This fixes the case of `example.com` which returns an HTML document with status 404, however sites that return their landing page or similar at `/.well-known/security.txt` with a 200 status code would still be parsed. 
Unsure, how to best handle such cases without introducing false negatives.

JannisBush:
@tunetheweb Can this be merged before the crawl starts tomorrow?

As written above there might still be a a very small number of sites with incorrect "other" values.
However, I think this does not pose a major problem:
- Such sites (Return an OK status code and show a non-security.txt file that contains lines that match our regex) should be rare and would only be a couple of text entries in an array -> no storage issues expected
- We save the content-type and maybe we can filter out incorrect sites using that 
- All other entries except for "Other" will be correct. 
- For "Other" the main analysis will probably be which other values occurred the most often, which is not affected by this problem

tunetheweb (approved):
LGTM

PR #116, opened 2024-04-08
Add custom metric listing the cookie store

rviscomi:
Progress on #112 

Adds a custom metric to dump the contents of the cookie jar

---

**Test websites**:

- https://example.com/
- https://web.dev/

PR #126, opened 2024-06-03
[Cookies Chapter 2024] Use newly exposed WPT_COOKIES in cookies.js

yohhaan:
Modification of the custom metric `cookies.js` to use the newly exposed `WPT_COOKIES`  which allows to directly access all cookies: 1st and 3rd parties, set through JS or HTTP.

This pull request documents the properties we would need for the cookies chapter for the Almanac 2024. See also my comment [here](https://github.com/HTTPArchive/custom-metrics/pull/116#issuecomment-2145568438) on  [pull request 116](https://github.com/HTTPArchive/custom-metrics/pull/116)

---
**Test websites**:

- https://example.com/
- https://web.dev/


rviscomi (approved):
@nrllh LGTY?

nrllh:
> @nrllh LGTY?

yes, it seems good.

PR #118, opened 2024-05-18
Privacy 2024

Yash-Vekaria:
Fixes/improves **ads.js**

Following changes are made:
1. Added extraction of count of confudential entries assoicated with each sellers.json file.
2. Handled crawling of sellers.json hosted on non-standard paths but connected via redirections.
3. Fixed blank sellers object in the output JSON by handling cases where domains or seller_type fields are absent.
4. Included collection of ad exchange and publisher domains for BQ analysis.
5. Google being popular ad-exchange hosting sellers.json at custom location, added its support.

Changes 1-4 are handled in ```Commit #fa17fc5d2c14ccedb5bca637dde45ef468ce52a6``` and Change 5 is handled in ```Commit #13c2040a7da673141177281a5fe1d4787fd52c3b```.
  
Testing was done on WPT Test for the following website and it worked fine.

---
**Test websites**:

- https://www.facebook.com
- https://www.amazon.com
- https://www.media.net

PR #127, opened 2024-06-03
[Cookies Chapter 2024] Adding Privacy Sandbox well-known files

yohhaan:
This pull request modifies the `well-known.js` custom metric to parse 2 well-known files related to Google's Privacy Sandbox:

- [related-website-set.json](https://github.com/GoogleChrome/related-website-sets)
- [privacy-sandbox-attestations.json](https://github.com/privacysandbox/attestation)

---

**Test websites**:

- https://mercadolibre.com
- https://www.media.net
- https://yu.ru

### DIFF
--- a/dist/well-known.js
+++ b/dist/well-known.js
@@ -3,7 +3,7 @@
 
 // README! Instructions for adding a new custom metric for the Web Almanac.
 //
-// 1. Refer for instructions for adding a custom metric in almanamc.js.
+// 1. Refer for instructions for adding a custom metric in almanac.js.
 // 2. This file has a special case where a custom metric uses 'fetch' and in that case we need to return a promise that resolves to JSON
 // 3. Test your change by following the instructions at https://github.com/HTTPArchive/almanac.httparchive.org/issues/33#issuecomment-502288773.
 // 4. Submit a PR to update this file.
@@ -83,6 +83,9 @@ return Promise.all([
   // ecommerce
   parseResponse('/.well-known/assetlinks.json'),
   parseResponse('/.well-known/apple-app-site-association'),
+  // privacy sandbox
+  parseResponse('/.well-known/related-website-set.json'), //Related Website Set
+  parseResponse('/.well-known/privacy-sandbox-attestations.json'), //Attestation File
   // privacy
   parseResponse('/.well-known/gpc.json', r => {
     return r.text().then(text => {


yohhaan:
I meant https://ya.ru/ for automated tests

SaptakS (approved):

tunetheweb (approved):
LGTM but you're not actually parsing the results and just checking the existence of the files. Is that all that you want to do as would have thought you might be ingerested in the contents of the files?

yohhaan:
Good point @tunetheweb

The current plan is to get a list of domains that potentially have one of these 2 files through the HTTP archive crawl.
Then, use a custom crawler (that I already have) to actually check if the detected JSON files are compliant with the expected JSON schemas, and then do some further analyses on the valid files.

I would be happy to do this parsing and JSON validation in the custom metric directly, but I would need to be able to call a JSON schema validator like the Ajv library and I am not sure how I would go about it. It is unclear to me if and how I can install further dependencies that these custom metrics would have access to (I am working on another metric where I would need the Public Suffix list to extract the eTLD+1 of the hostnames).

max-ostapenko:
@yohhaan I had an idea from the documentation that Chrome will consume this [list of submitted and validated domains](https://github.com/GoogleChrome/related-website-sets/blob/main/related_website_sets.JSON).
So I thought it doesn't require additional scanning, no?

max-ostapenko:
FYI Privacy Sandbox attestations implemented as a simple check [within Privacy chapter PR](https://github.com/HTTPArchive/custom-metrics/pull/118/files#:~:text=const-,attestationPublished,-%3D%20await%20fetchAndCheckAttestation), as part of `privacy-sandbox` custom metric.

yohhaan:
@max-ostapenko:

**Related Website Set:** this is indeed supposed to be the "canonical" list consumed by Chrome, but in practice some domains are not listed there... 
As an example: https://google.com/.well-known/related-website-set.json

I would like to get from the HTTP Archive crawl viewpoint which websites may have this file set, and then do further post-analysis to check if they are in that "canonical" list or no, if the file they host is exactly the same as the one published, etc.

**Attestation:** yes, I saw the privacy chapter PR (I am currently working on adding detection of other Privacy Sandbox APIs based on the proposed `privacy-sandbox.js` metric by @Yash-Vekaria ).

Checking in `well-known.js` and `privacy-sandbox.js` would actually be complementary as I see it: 
- in `well-known.js` we check for if the file exists on the origin crawled (with the expectation that this file will not exist most of the time, except if we are crawling the homepage of an advertiser, etc., but we may discover it on unexpected websites)
- in `privacy-sandbox.js`, we check for the third parties that we detect calling the Privacy Sandbox APIs, if they have the attestation file set (as they should).

PR #128, opened 2024-06-06
Added custom metric for ads.js

Yash-Vekaria:
Fixes/improves **ads.js**

Following changes are made:
1. Added extraction of count of confidential entries assoicated with each sellers.json file.
2. Handled crawling of sellers.json hosted on non-standard paths but connected via redirections.
3. Fixed blank sellers object in the output JSON by handling cases where domains or seller_type fields are absent.
4. Included collection of ad exchange and publisher domains for BQ analysis.
  
Testing was done on WPT Test for the following website and it worked fine. The changes were reviewed in this PR: [#118](https://github.com/HTTPArchive/custom-metrics/pull/118) 

---
**Test websites**:

- https://www.facebook.com
- https://www.amazon.com
- https://www.media.net

### DIFF
--- a/dist/ads.js
+++ b/dist/ads.js
@@ -6,9 +6,15 @@ const SELLER_TYPES = ['publisher', 'intermediary', 'both'];
 const isPresent = (response, endings) => response.ok && endings.some(ending => response.url.endsWith(ending));
 
 const fetchAndParse = async (url, parser) => {
+  const timeout = 5000;
+  /*
+  Google's sellers.json size is 120Mb as of May 2024 - too big for custom metrics.
+  It's available at realtimebidding.google.com/sellers.json, so not part of crawled pages list.
+  More details: https://support.google.com/authorizedbuyers/answer/9895942
+  */
   const controller = new AbortController();
   const { signal } = controller;
-  setTimeout(() => controller.abort(), 5000);
+  setTimeout(() => controller.abort(), timeout);
 
   try {
     const response = await fetch(url, { signal });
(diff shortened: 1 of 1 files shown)


dist/ads.js
```diff
@@ -84,9 +90,9 @@ const parseAdsTxt = async (response) => {
     // Count unique and remove domain Sets for now
     for (let accountType of Object.values(result.account_types)) {
       accountType.domain_count = accountType.domains.size;
-      delete accountType.domains // Keeping a list of domains may be valuable for further research, e.g. accountType.domains = [...accountType.domains];
+      accountType.domains = Array.from(accountType.domains); // delete accountType.domains
```
tunetheweb, line 93: @max-ostapenko why did you delete this before despite saying it might be valuable? Is it OK to add them back in?

max-ostapenko: @tunetheweb It adds quite some data, so I expected to bring it as a second step.
And now we have a clear use case for this data in WebAlmanac - exploring ads supply  transparency and transparency.

tunetheweb: Hmmm I see what you mean. That is quite a bit of data. WDYT @rviscomi ?

rviscomi: If the data is useful, LGTM. I'd imagine this only affects a fraction of sites in the dataset, but let's run it for one month and roll it back if we see it significantly affects the size of the crawl.

tunetheweb (approved):
LGTM

PR #129, opened 2024-06-06
Adding Privacy Sandbox APIs

Yash-Vekaria:
Implemented capturing the following Privacy Sandbox APIs:
1.  Attribution Reporting API
2. Federated Credentials Manager API
3. Fenced Frames API
4. FLoC API
5. Protected Audience API
6. Private Aggregation API
7. Private State Tokens API
8. Shared Storage API
9. Storage Access API
10. Topics API
11. User Client Hints API

---
**Test websites**:

- https://www.operafootball.com/
- https://floc.glitch.me/
- https://pets-animals-pets-cats.glitch.me/
- https://tennis-tennis.glitch.me/
- https://protected-audience-demo.web.app/
- https://protected-audience-demo-advertiser.web.app/
- https://protected-audience-demo-publisher.web.app/
- https://fedcm-rp-demo.glitch.me/
- https://fedcm-idp-demo.glitch.me/
- https://private-state-token-redeemer.glitch.me/
- https://private-state-token-issuer.glitch.me/
- https://user-agent-client-hints.glitch.me/
- https://shared-storage-demo.web.app/
- https://www.example.com/


dist/privacy-sandbox.js
```diff
@@ -0,0 +1,402 @@
...
+     **************************************************************************/
+
+    //Todo
+
+    /***************************************************************************
+     * FLoC (deprecated - are some still calling it?))
+     * Documentation: https://web.dev/articles/floc
+     * Test site: https://floc.glitch.me/
+     **************************************************************************/
+    if (checkResponseBody(request, 'document.interestCohort')) {
```
max-ostapenko, line 238: @yohhaan [Implemented currently](https://github.com/HTTPArchive/custom-metrics/blob/fc9f4f4f7cf7220164c556216b2cf067631887ed/dist/privacy.js#L178). But planned for deprecation in [this PR](https://github.com/HTTPArchive/custom-metrics/pull/117).
Do you want to leave it moving forward, or historical data is enough?

tunetheweb:
@Yash-Vekaria @max-ostapenko what's the latest on this PR as need to merge it today if we want it included in the Web Almanac 2024.

Yash-Vekaria:
> @Yash-Vekaria @max-ostapenko what's the latest on this PR as need to merge it today if we want it included in the Web Almanac 2024.

@tunetheweb Topics API is ready to merge. Protected Audience API and Attribution APIs have been correctly implemented but I don't have a good example that simulates all the cases, so partial response was recorded on the current test website. I'll let @yohhaan comment if he plans to incorporate any other APIs. @max-ostapenko informed me that he shall be taking a look at the code later today to finalize these APIs and perhaps remove stale code to make the PR ready to merge.

max-ostapenko:
@tunetheweb let's merge it.
I'll make my PR on top of main then

yohhaan:
Hi,
I am modifying at the moment the Topics API implementation (some errors in the current one), I will push shortly and add more details about the rest.

dist/privacy-sandbox.js
yohhaan: Even though the API is deprecated it could be nice to leave the detection on to see if code not up to date is still calling it.

yohhaan:
Can we add the following test sites to this PR for unit tests (I am still checking if header detection is working as expected)?
- https://floc.glitch.me/   (FLoC)
- https://pets-animals-pets-cats.glitch.me/  (Topics)
- https://tennis-tennis.glitch.me/   (Topics)
- https://protected-audience-demo.web.app/ (Protected Audience)
- https://protected-audience-demo-advertiser.web.app/ (Protected Audience)
- https://protected-audience-demo-publisher.web.app/ (Protected Audience)
 
 Also, I am not sure the async check for the attestation is working/returning

dist/privacy-sandbox.js
max-ostapenko: @yohhaan Do you want to keep page level metric as it was implemented in privacy.js for 2024 analysis?

yohhaan: You can deprecate the page level metric, and we can just keep the detection in privacy-sandbox.js

dist/privacy-sandbox.js
```diff
@@ -0,0 +1,502 @@
...
+ *
+ * Required command line flags: --enable-features=BrowsingTopics,InterestGroupStorage,PrivacySandboxAdsAPIsOverride
+ *
+ * Documentation Permissions Policy: https://developers.google.com/privacy-sandbox/relevance/setup/web/permissions-policy
+*/
+
+let requests = $WPT_BODIES;
+
+let result = { // alphabetical order for organization
+  apiCallersAttestation: {}, // object {origin (string): true/false} Note: true does not especially mean that attestation file is valid (need to check if JSON, compliant with JSON schema, etc.)
```
tunetheweb, line 20:
```suggestion
  apiCallersAttestation: {}, // object {domain (string): true/false} Note: true does not especially mean that attestation file is valid (need to check if JSON, compliant with JSON schema, etc.)
```

dist/privacy-sandbox.js
```diff
@@ -0,0 +1,502 @@
...
+ * @function apiCallerAdd
+ * Add provided caller API origin to results if not already present
+ *
+ * @param {string} origin - API caller origin
+ */
+function apiCallerAdd(origin) {
+  if (!(origin in result.apiCallersAttestation)) {
+    result.apiCallersAttestation[origin] = null;
+  }
+}
```
tunetheweb, line 168:
```suggestion
 * Add provided caller API domain to results if not already present
 *
 * @param {string} domain - API caller domain
 */
function apiCallerAdd(domain) {
  if (!(domain in result.apiCallersAttestation)) {
    result.apiCallersAttestation[domain] = null;
  }
}
```

dist/privacy-sandbox.js
```diff
@@ -0,0 +1,502 @@
...
+ * they call 1 API they are likely to call others too and so have that file published)
+ *
+ * @param {string} origin - API caller origin
+ */
+async function fetchAttestations() {
+  for (const [origin, _] of Object.entries(result.apiCallersAttestation)) {
+    const attestation = fetchAndCheckResponse(`${'https://' + origin}/.well-known/privacy-sandbox-attestations.json`);
+    result.apiCallersAttestation[origin] = attestation;
+  }
+}
```
tunetheweb, line 188:
```suggestion
 * Attestation is supposed to be under same https://domain than API caller, no redirect
 * https://github.com/privacysandbox/attestation
 * only required for attribution reporting, topics, protected audience, shared
 * storage, and private aggregation APIs so far
 * we check anyway if response returned for any caller of any API we detect (if
 * they call 1 API they are likely to call others too and so have that file published)
 *
 * @param {string} domain - API caller domain
 */
async function fetchAttestations() {
  for (const [domain, _] of Object.entries(result.apiCallersAttestation)) {
    const attestation = fetchAndCheckResponse(`${'https://' + domain}/.well-known/privacy-sandbox-attestations.json`);
    result.apiCallersAttestation[domain] = attestation;
  }
}
```

dist/privacy-sandbox.js
```diff
@@ -0,0 +1,502 @@
...
+      // [XHR] '{deprecatedBrowsingTopics: true}' (to be deprecated)
+      if (respHeaders.has('observe-browsing-topics') && respHeaders.get('observe-browsing-topics') === "?1") {
+        // [response header] 'Observe-Browsing-Topics: ?1' to include page in topics calculation
+        result['topicsAPI']['browsingTopicsJs'].push({ "origin": requestDomain, "skipObservation": false });
+        apiCallerAdd(requestDomain);
+      } else {
+        result['topicsAPI']['browsingTopicsJs'].push({ "origin": requestDomain, "skipObservation": true });
+        apiCallerAdd(requestDomain);
+      }
+    }
```
tunetheweb, line 461:
```suggestion
    if (checkResponseBody(request, 'document.browsingTopics\(\s*\)')) {
      // [javascript] 'document.browsingTopics()'
      result['topicsAPI']['browsingTopicsJs'].push({ "domain": requestDomain, "skipObservation": false });
      apiCallerAdd(requestDomain);
    }
    if (checkResponseBody(request, 'document.browsingTopics\(\s*\{\s*skipObservation\s*:\s*true\s*\}\s*\)')) {
      // [javascript] 'document.browsingTopics({skipObservation:true})'
      result['topicsAPI']['browsingTopicsJs'].push({ "domain": requestDomain, "skipObservation": true });
      apiCallerAdd(requestDomain);
    }
    if (checkResponseBody(request, '\{\s*browsingTopics\s*:\s*true\s*\}') || checkResponseBody(request, '\{\s*deprecatedBrowsingTopics\s*:\s*true\s*\}')) {
      // [fetch] '{browsingTopics: true}'
      // [XHR] '{deprecatedBrowsingTopics: true}' (to be deprecated)
      if (respHeaders.has('observe-browsing-topics') && respHeaders.get('observe-browsing-topics') === "?1") {
        // [response header] 'Observe-Browsing-Topics: ?1' to include page in topics calculation
        result['topicsAPI']['browsingTopicsJs'].push({ "domain": requestDomain, "skipObservation": false });
        apiCallerAdd(requestDomain);
      } else {
        result['topicsAPI']['browsingTopicsJs'].push({ "domain": requestDomain, "skipObservation": true });
        apiCallerAdd(requestDomain);
      }
    }
```

dist/privacy-sandbox.js
```diff
@@ -0,0 +1,502 @@
...
+      // [request header] 'Sec-Browsing-Topics: true'
+      if (respHeaders.has('observe-browsing-topics') && respHeaders.get('observe-browsing-topics') === "?1") {
+        // [response header] 'Observe-Browsing-Topics: ?1' to include page in topics calculation
+        result['topicsAPI']['browsingTopicsHeader'].push({ "origin": requestDomain, "skipObservation": false });
+        apiCallerAdd(requestDomain);
+      } else {
+        result['topicsAPI']['browsingTopicsHeader'].push({ "origin": requestDomain, "skipObservation": true });
+        apiCallerAdd(requestDomain);
+      }
+    }
```
tunetheweb, line 473:
```suggestion
    if (reqHeaders.has('sec-browsing-topics')) {
      // [request header] 'Sec-Browsing-Topics: true'
      if (respHeaders.has('observe-browsing-topics') && respHeaders.get('observe-browsing-topics') === "?1") {
        // [response header] 'Observe-Browsing-Topics: ?1' to include page in topics calculation
        result['topicsAPI']['browsingTopicsHeader'].push({ "domain": requestDomain, "skipObservation": false });
        apiCallerAdd(requestDomain);
      } else {
        result['topicsAPI']['browsingTopicsHeader'].push({ "domain": requestDomain, "skipObservation": true });
        apiCallerAdd(requestDomain);
      }
    }
```

dist/privacy-sandbox.js
```diff
@@ -0,0 +1,502 @@
...
+
+    if (checkResponseBody(request, 'document.requestStorageAccess\(')) {
+      // [javascript] document.requestStorageAccess(types: optional)
+      result['relatedWebsiteSet']['requestStorageAccess'].push(requestDomain);
+      apiCallerAdd(requestDomain);
+    }
+
+    if (checkResponseBody(request, 'document.requestStorageAccessFor\(')) {
+      // [javascript] document.requestStorageAccessFor(requestedOrigin)
+      result['relatedWebsiteSet']['requestStorageAccessFor'].push(requestDomain);
```
tunetheweb, line 371: Is this origin or domain?

tunetheweb (changes requested):
There's a lot of confusion of origins and domains in this. Suggested some clean ups.

dist/privacy-sandbox.js
yohhaan: The API call takes an origin as argument, but we are logging the domain from which the request making the call is coming from. So, `result['relatedWebsiteSet']['requestStorageAccessFor'].push(requestDomain);`

tunetheweb: OK makes sense.

tunetheweb (approved):
LGTM.

@max-ostapenko any further comments or you happy to merge?

tunetheweb:
This sort of thing is very verbose and needlessly increases the size of custom metrics for all 15 million sites (most of which likely don't use shared storage):

```
        "sharedStorage": {
            "shared-storage": true,
            "shared-storage-select-url": true,
            "append": [],
            "clear": [],
            "delete": [],
            "set": [],
            "run": [],
            "selectURL": [],
            "addModule": []
        },
```

Can we exclude empty values in all the data in this custom metric?

I made a similar comment on https://github.com/HTTPArchive/custom-metrics/pull/125#pullrequestreview-2097390279 because if EVERY custom metric adds this sort of thing then the custom metric (already 8.32TB or $50 for a single query) will become too large for anyone to query it without taking out a mortgage.

yohhaan:
Sounds good, on it

> This sort of thing is very verbose and needlessly increases the size of custom metrics for all 15 million sites (most of which likely don't use shared storage):
> 
> ```
>         "sharedStorage": {
>             "shared-storage": true,
>             "shared-storage-select-url": true,
>             "append": [],
>             "clear": [],
>             "delete": [],
>             "set": [],
>             "run": [],
>             "selectURL": [],
>             "addModule": []
>         },
> ```
> 
> Can we exclude empty values in all the data in this custom metric?
> 
> I made a similar comment on [#125 (review)](https://github.com/HTTPArchive/custom-metrics/pull/125#pullrequestreview-2097390279) because if EVERY custom metric adds this sort of thing then the custom metric (already 8.32TB or $50 for a single query) will become too large for anyone to query it without taking out a mortgage.

yohhaan:
Per @tunetheweb 's comment: I changed the format of the result object to make it smaller if APIs are not called.

@Yash-Vekaria: can you see if you can port the  ProtectedAudience and Attribution Reporting APIs to a similar format?

max-ostapenko:
@yohhaan I couldn't find any way to verify cross-origin files existence.
Fetch doesn't fail because of 404.
Do you have any ideas?

Otherwise I would say we remove the attestation checks from custom metrics completely.

tunetheweb:
Same comment, do we need these all to be set to empty values or can we skip them when nothing of interest in them to save storage space/query costs?

```
        "fencedFrames": [],
        "floc": [],
        "privateAggregation": [],
        "privateStateTokens": [],
        "protectedAudienceAPI": {
            "interestGroups": {
                "joinAdInterestGroup": [],
                "leaveAdInterestGroup": [],
                "updateAdInterestGroups": [],
                "clearOriginJoinedAdInterestGroups": []
            },
            "runAdAuction": [],
            "generateBid": [],
            "scoreAd": [],
            "reportWin": [],
            "reportResult": []
        },
        "sharedStorage": [],
        "storageAccess": [],
        "topicsAPI": [],
        "userAgentClientHints": []
```

yohhaan:
I am done with the edits I wanted to do on this custom metric, the tests have started to return results as expected since I fixed the double escape issue for the regex.

If people have more cycles to commit, they could try to find websites to test on for each API and see that the calls are actually correctly detected. I have done a few tests + the ones from the automated GitHub actions and things look good to me for now.
Also, note: some websites listed to test on require user interaction, thus the empty results sometimes.

Thanks to @Yash-Vekaria for all the help along the implementation of this custom metric and to @max-ostapenko as well!

Yash-Vekaria:
Thanks @yohhaan and @max-ostapenko for bringing this to completion.

@tunetheweb All the changes are locked from my side as well. **We are ready to merge this PR**.

@max-ostapenko Extraction of attestation has been removed as it was working in WPT without Privacy Sandbox API capturing in WPT but not with it on live checks (some async/promise issue thats hard to debug due to different behavior). We have decided to analyse it in analysis phase and remove it in custom metrics altogether.

tunetheweb:
I still think that is a lot of extra data to store per URL when in a lot of cases there won’t be any different. Is there no way to only report if the defaults are different?

I tried adding https://www.example.com to see what it would produce and reran the test but it didn’t pick them up. But I’m presuming it would produce the same as the blank ones?

Yash-Vekaria:
@tunetheweb One way to optimize further is that we remove all `permission-policy` flags (as shown below) because now that I'm thinking more about it, they should always be `true` (**except** for "_top-level-storage-access_"). Using command line activation fo Privacy Sandbox APIs for the browser being used makes these set to `true` by default.
```
"permissionsPolicy": {
            "attribution-reporting": true,
            "browsing-topics": true,
            "identity-credentials-get": true,
            "interest-cohort": true,
            "join-ad-interest-group": true,
            "private-aggregation": true,
            "run-ad-auction": true,
            "shared-storage": true,
            "shared-storage-select-url": true,
            "storage-access": true,
            "top-level-storage-access": false
        },
```

Next, for other APIs, we can have a domain-level mapping where key is domain and the value is a list of APIs used by that domain on the current website, rather than having separate keys for each API and same domain repeating in each of them.

So, reduced output will look like the following for operafootball.com:
```
{
    "_privacy-sandbox": {
        "top-level-storage-access": false
        "securepubads.g.doubleclick.net": ["runAdAuction", "getHighEntropyValues"],
        "pagead2.googlesyndication.com": ["runAdAuction", "joinAdInterestGroup", "getHighEntropyValues"],
        "www.googletagmanager.com": ["joinAdInterestGroup", "getHighEntropyValues"],
        "googleads.g.doubleclick.net": ["browsingTopicsHeader|true",],
    }
}
```

tunetheweb:
That looks a lot better if you can do that. And happy to include permission policy data if it;'s different from the default (i.e. explicitly blocked) but seems silly to log that same default of all true millions of times.

Yash-Vekaria:
@tunetheweb I have refactored the code as shared in the previous comment. If it looks good, we can merge.

tunetheweb:
Thanks for that. Merging!

PR #130, opened 2024-06-12
Add try/catch to failing privacy test suite

tunetheweb:
Fixed issue preventing privacy CS from running on many sites

---

**Test websites**:

- https://example.com/


pmeenan (approved):
LGTM

PR #132, opened 2024-06-14
Log custom metrics for webalmanac

max-ostapenko:
Now you can see custom metric logged by default from running unit tests with Web Almanac.

Applied formatting to `privacy.js` to trigger the 'changed custom metric' test.

---
**Test websites**:

- https://example.com/


tunetheweb (approved):
LGTM.

Will wait until crawl finished before merging though just in case!

PR #133, opened 2024-06-18
fix imagesrcset bug

kara:
28% of Angular 17 apps using the image directive have been missing performance data due to errors. Looking into it, it seems that `link.imagesrcset` is undefined in some of these cases. Grabbing the attribute instead of the prop with `link.getAttribute` fixes the issue.

See example failure: https://www.webpagetest.org/result/240618_AiDc8Y_E9D/3/details/
See example fix: https://www.webpagetest.org/result/240618_AiDcZ3_E78/2/details/

---
**Test websites**:

- https://exploreti.com/

### DIFF
--- a/dist/performance.js
+++ b/dist/performance.js
@@ -264,7 +264,7 @@ function getLcpPreloadInfo(rawDoc, lcpUrl) {
 
         let src = link.href;
         if (link.hasAttribute('imagesrcset')) {
-            src = splitSrcSet(link.imagesrcset).find(src => src == lcpUrl);
+            src = splitSrcSet(link.imagesrcset || link.getAttribute('imagesrcset')).find(src => src == lcpUrl);
         }
 
         return src == lcpUrl;
--- a/dist/performance.js
+++ b/dist/performance.js
@@ -264,7 +264,7 @@ function getLcpPreloadInfo(rawDoc, lcpUrl) {
 
         let src = link.href;
         if (link.hasAttribute('imagesrcset')) {
-            src = splitSrcSet(link.imagesrcset || link.getAttribute('imagesrcset'))?.find(src => src == lcpUrl);
+            src = splitSrcSet(link.getAttribute('imagesrcset'))?.find(src => src == lcpUrl);
         }
 
         return src == lcpUrl;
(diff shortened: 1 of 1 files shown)


dist/performance.js
```diff
@@ -264,7 +264,7 @@ function getLcpPreloadInfo(rawDoc, lcpUrl) {
 
         let src = link.href;
         if (link.hasAttribute('imagesrcset')) {
-            src = splitSrcSet(link.imagesrcset).find(src => src == lcpUrl);
+            src = splitSrcSet(link.imagesrcset || link.getAttribute('imagesrcset')).find(src => src == lcpUrl);
```
tunetheweb, line 267: WDYT about adding an optional chaining operator as an extra check?

```suggestion
            src = splitSrcSet(link.imagesrcset || link.getAttribute('imagesrcset'))?.find(src => src == lcpUrl);
```

kara: SGTM, I was debating whether that was necessary

tunetheweb: Technically, it shouldn't be!

dist/performance.js
```diff
@@ -264,7 +264,7 @@ function getLcpPreloadInfo(rawDoc, lcpUrl) {
 
         let src = link.href;
         if (link.hasAttribute('imagesrcset')) {
-            src = splitSrcSet(link.imagesrcset).find(src => src == lcpUrl);
+            src = splitSrcSet(link.imagesrcset || link.getAttribute('imagesrcset'))?.find(src => src == lcpUrl);
```
tunetheweb, line 267: Actually I've found the bug!!! That's to [Jake's amazing post](https://jakearchibald.com/2024/attributes-vs-properties/) (thank you @jakearchibald!). The property is case sensitive and has the incorrect casing.

So this works (with optional chaining _just_ in case):

```suggestion
            src = splitSrcSet(link.imageSrcset)?.find(src => src == lcpUrl);
```

tunetheweb: Though maybe get attribute is safer in that case as not 100% obvious in case someone corrects it later for some reason? But don't think we need both:

```suggestion
            src = splitSrcSet(ink.getAttribute('imagesrcset'))?.find(src => src == lcpUrl);
```

kara: LOL that casing is so unintuitive! +1 to using the attribute

tunetheweb (approved):

PR #134, opened 2024-06-18
Fix performance custom metrics

tunetheweb:
Noticed performance custom metrics were not getting logged in our CI tests.

This was because the current logic was as follows:

```js
          try {
            wpt_custom_metrics[`_${metric_name}`] = JSON.parse(wpt_custom_metric);
            if (metrics_to_log.includes(metric_name)) {
              wpt_custom_metrics_to_log[`_${metric_name}`] = JSON.parse(wpt_custom_metric);
            }

          } catch (e) {
            wpt_custom_metrics[`_${metric_name}`] = wpt_custom_metric;
          }
```

That says only log metrics we're interested in if `JSON.parse` succeeds.

However, many metrics are returned as Objects rather than JSON and so `JSON.parse` fails (as it doesn't need to be parsed as it'a already an object!), and others aren't valid JSON as just plain strings. In both cases the `if (metrics_to_log.includes(metric_name)) {` part isn't called and they are excluded from the output.

This PR refactors to move code out of the `try` to log in both cases.

---

**Test websites**:

- https://example.com/
- https://exploreti.com/

### DIFF
--- a/dist/performance.js
+++ b/dist/performance.js
@@ -371,3 +371,4 @@ return Promise.all([getLcpElement()]).then(([lcp_elem_stats]) => {
 }).catch(error => {
     return {error};
 });
+
--- a/dist/performance.js
+++ b/dist/performance.js
@@ -371,4 +371,3 @@ return Promise.all([getLcpElement()]).then(([lcp_elem_stats]) => {
 }).catch(error => {
     return {error};
 });
-
--- a/tests/wpt.js
+++ b/tests/wpt.js
@@ -73,9 +73,9 @@ function runWPTTest(url) {
         for (const metric_name of custom_metrics) {
           wpt_custom_metric = response.data.runs['1'].firstView[`_${metric_name}`];
           try {
-            wpt_custom_metrics[`_${metric_name}`] = JSON.parse(wpt_custom_metric);
+            wpt_custom_metrics[`_${metric_name}`] = JSON.parse(wpt_custom_metric.toString());
             if (metrics_to_log.includes(metric_name)) {
-              wpt_custom_metrics_to_log[`_${metric_name}`] = JSON.parse(wpt_custom_metric);
+              wpt_custom_metrics_to_log[`_${metric_name}`] = JSON.parse(wpt_custom_metric.toString());
             }
 
           } catch (e) {
--- a/tests/wpt.js
+++ b/tests/wpt.js
@@ -79,7 +79,7 @@ function runWPTTest(url) {
             }
 
           } catch (e) {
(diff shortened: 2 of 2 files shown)


max-ostapenko:
Thanks @tunetheweb 

Do you know why are some metrics wrapped into JSON.stringify and others not?

max-ostapenko (approved):

tunetheweb:
> Thanks @tunetheweb
> 
> Do you know why some metrics are wrapped into JSON.stringify and others not? Would anything change in BQ if we drop JSON.stringify?

Some of the custom metrics call `JSON.stringify()` in the return ([example for media custom-metric](https://github.com/HTTPArchive/custom-metrics/blob/b498ac11790b6d93ee89a3516a434ccf4aff8acc/dist/media.js#L49)) and some don't ([example for performance custom-metric](https://github.com/HTTPArchive/custom-metrics/blob/b498ac11790b6d93ee89a3516a434ccf4aff8acc/dist/performance.js#L360) which just returns the object directly).

Mostly it doesn't matter, and Web Page Test will handle both and `stringify()` automatically when it needs to store them (e.g. when saving to it's Database).

However, one place it CAN matter is if you're running the code in the console to test it. You can get unexpected results if you're not `stringify`ing it. See long conversation on this here: https://github.com/HTTPArchive/custom-metrics/pull/113#issuecomment-2043937823

Probably we should standardise this and update all the custom metrics to `JSON.stringify()` the final `return` so we don't have this inconsistency. But in meantime this works.

tunetheweb:
Updated PR to add the comment to explain why.

max-ostapenko:
> Probably we should standardise this and update all the custom metrics to `JSON.stringify()` the final `return` so we don't have this inconsistency. But in meantime this works.

@tunetheweb so maybe we rather wrap performance CM (and other metrics that have the same issue) in `JSON.stringify`?

tunetheweb:
Yeah we probably should. But I think we'd still want to keep this change in anyway incase a future metric was added without it, of some of the non-JSON custom metrics can also return invalid JSON (and converting those to JSON is a bigger job with bigger implications).

Plus it's more efficient (calls `JSON.parse` only once).

So not super urgent IMHO once we merge this.

max-ostapenko:
I think this PR makes these tests more similar to the events processing logic.

tunetheweb:
> I think this PR makes these tests more similar to the events processing logic.

Not sure what you mean? And if you are saying that was a good thing or a bad thing 😀

tunetheweb:
Oh very similar! Nice find. OK merging this.

PR #135, opened 2024-07-04
Add File System Observer

tomayac:
Add File System Observer custom metric check


max-ostapenko:
Reason for test error - the WPT test request needs to be forced to run as POST.
Adjusted in [#136 ](https://github.com/HTTPArchive/custom-metrics/pull/136/files#diff-43f0d33cb7f304fea53c3c82c2f410198265e3f591512a9e027ac76d277737cbR88)

PR #136, opened 2024-07-07
Ads files redirected

max-ostapenko:
Sometimes `sellers.json` redirects to the main adtech domain (for some business groups or managed domains), and the original page would be listed as a seller/publisher page on the file. Need to use the domain from redirect URL, instead of `page`.
- [x] tracking redirect final url

Plus several updates to the test action:
- [x] forcing WPT requests to use POST method to send all the custom metrics code (fixes 'Too long URL' error),
- [x] in some cases the length of a PR comment content is too long (65K chars limit), so saving .md file as artifact instead,
- [x] moved most of the code out of yaml to simplify the workflow update process

As the changes are not yet applied in test check - it fails. See https://github.com/max-ostapenko/custom-metrics/pull/1 instead.

---
**Test websites**:

- https://www.example.com/

### DIFF
--- a/.github/workflows/wpt-test.yml
+++ b/.github/workflows/wpt-test.yml
@@ -47,8 +47,9 @@ jobs:
             artifact_url="${{ steps.artifact-upload-step.outputs.artifact-url }}"
             sed -i "s|{artifact-url}|$artifact_url|g" comment.md
             echo "Placeholder {artifact-url} replaced with a value: $artifact_url."
+          else
+            echo "No placeholder {artifact-url} found in comment.md."
           fi
-
 
       - name: Add comment to PR
         uses: mshick/add-pr-comment@v2
--- a/.github/workflows/wpt-test.yml
+++ b/.github/workflows/wpt-test.yml
@@ -29,7 +29,6 @@ jobs:
           WPT_API_KEY: ${{ secrets.HA_API_KEY }}
 
       - name: Run WebPageTest for more websites
-        id: websites-tests
         run: node tests/wpt.js
         env:
           WPT_SERVER: "webpagetest.httparchive.org"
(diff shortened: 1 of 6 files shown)


max-ostapenko:
@tunetheweb just want to clarify that regardless of failing checks, it's ready for review and hopefully merge.

rviscomi (approved):

.github/workflows/wpt-test.yml
```diff
@@ -31,57 +29,31 @@ jobs:
...
-          # If the index is valid, then parse the URLs
-          if [ $start_index -gt -1 ]; then
-            # Initialize an array for URLs
-            declare -a URLS
-            url_pattern="((http|https|ftp):\/\/[a-zA-Z0-9.-]+(\.[a-zA-Z]{2,4})(\/[a-zA-Z0-9_.-]+)*(\/?)(\?[a-zA-Z0-9_.-]+=[a-zA-Z0-9%_.-]+)*(\#?)([a-zA-Z0-9%_.-=]+)*)"
+        run: node tests/wpt.js
+        env:
+          WPT_SERVER: "webpagetest.httparchive.org"
+          WPT_API_KEY: ${{ secrets.HA_API_KEY }}
+          PR_BODY: ${{ github.event.pull_request.body }}
```
rviscomi, line 36: What's weird about [this test run](https://github.com/HTTPArchive/custom-metrics/actions/runs/10047940354/job/27803652669?pr=137) is that it's using the original PR description before it was edited. There was a ~15 hour gap between the edit and the test run, so I'm not sure why it's not picking up the test websites that were added. Any ideas?

max-ostapenko: I see you triggered to re-run the previously run action here - it uses the same inputs as the initial attempt. I'm not 100% sure, but that's what I always experienced before.

The easiest way here would be to do a new commit on the PR branch and check the newly triggered run.

rviscomi: Ok thanks, makes sense

PR #140, opened 2024-08-20
Add dom-shape metric to markup

tunetheweb:
Adds a ["dom shape"](https://github.com/1cg/shape-of-the-web/blob/main/shape.js) metric as requested by @1cg

---

**Test websites**:

- https://example.com/
- https://www.cnn.com/
- https://web.dev/

### DIFF
--- a/dist/markup.js
+++ b/dist/markup.js
@@ -645,6 +645,146 @@ try { // whole process is placed in a try/catch so we can log uncaught errors
       return {
         hrefs_without_special_scheme
       };
+    })(),
+
+    'dom-shape': (() => {
+
+      //================================================================
+      //  helpers
+      //================================================================
+
+      function depthOf(node, currentDepth = 0) {
+          if (node === document.body) {
+              return currentDepth;
+          } else {
+              return depthOf(node.parentNode, currentDepth + 1);
+          }
+      }
+
+      function mean(arr) {
+          let total = arr.reduce((acc, curr) => acc + curr, 0)
+          return total / arr.length;
+      }
+
+      function stddev(arr, mean) {
+          let squareDiffs = arr.map((k) => (k - mean) ** 2);
+          let sumSquareDiffs = squareDiffs.reduce((acc, curr) => acc + curr, 0);
+          return Math.sqrt(sumSquareDiffs / arr.length);
+      }
+
+      function skew(mean, mode, stdDev) {
+          return (mean - mode) / stdDev;
+      }
+
+      function median(sortedArr) {
+          let length = sortedArr.length;
+          if (length % 2 === 0) {
+              return (sortedArr[length / 2 - 1] + sortedArr[length / 2]) / 2;
+          } else {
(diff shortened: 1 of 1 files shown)


tunetheweb:
Hi @1cg, I've added this custom metric as provided (except I put the "helper functions" to the top).

It seems to be working but let me know if there's some other site you want to test this on.

Some observations/questions from me:
- Should the `branchFactorByDepth` show the depth as a property?
- The `branchFactorByDepth` can get quite large for some sites (look at the cnn example). Is all this data needed?
- Are you wanting this data long term (i.e. we should leave this to run ever month), or only as a once off (in which case we ran remove it after the next crawl)?

After you answer those and confirm this is how you want the data we can get some of the other maintainers (@rviscomi , @pmeenan , @paulcalvano ) to weigh in with their thoughts.

1cg:
Hi @tunetheweb thank you very much for including this.  To answer your questions:

* The depth of the `branchFactorByDepth` array is implied by the index, but if you think that it would make sense to add it to make querying easier that would be easy to do, just let me know.
* Yes, in order to calculate the "shape" of the page we do need all that data.  However we could do something less computationally and storage intensive (e.g. just node count per depth and then estimating means, although that would perclude calculating skew, range, etc. for fanout)
* This can be a once off run, although doing it yearly might be nice to update the paper we are going to produce based on this data

Another thing that would be super useful would be metadata about the sites: total traffic, category, etc. so we can do statistical analysis across those sorts of variables.

I wanted to note that you and I have had off-github discussions about the performance of this analysis: if it is too heavy I am happy to dial it back.

Thank you!

tunetheweb:
> Yes, in order to calculate the "shape" of the page we do need all that data. However we could do something less computationally and storage intensive (e.g. just node count per depth and then estimating means, although that would perclude calculating skew, range, etc. for fanout)

OK let's leave as is. It's not that much data compared to other stuff.

> This can be a once off run, although doing it yearly might be nice to update the paper we are going to produce based on this data

Let's add and leave it to run monthly for now as it does fit in with [our mission](https://httparchive.org/about#mission). Can remove again if it proves less useful than we hoped.

> Another thing that would be super useful would be metadata about the sites: total traffic, category, etc. so we can do statistical analysis across those sorts of variables.

For "total traffic" we have [rank](https://developer.chrome.com/blog/crux-rank-magnitude) which is a broad measure of popularity in poer of 10 buckets, with half step intervals (so top 1k, top 5k, top 10k, top 50k, top 100k...etc.).

For category, we recently added [classification via the Topics API](https://github.com/HTTPArchive/httparchive.org/issues/868), however some have noted accuracy issues with that classificaiton with this once you get out of the most popular pages.

> I wanted to note that you and I have had off-github discussions about the performance of this analysis: if it is too heavy I am happy to dial it back.

My tests seem to show it's super quick and had a quick discussion with the other maintainers and they don't seem concerned either.

pmeenan (approved):
LGTM

PR #141, opened 2024-08-25
Update a11y.js for 2025

mgifford:
Resolves \#1

This will expand it to include Drupal 7 and a little CMS called WordPress that might have a bit of traction.

---
**Test websites**:

- https://wordpress.org ?
- https://backdropcms.org ?

### DIFF
--- a/dist/a11y.js
+++ b/dist/a11y.js
@@ -215,7 +215,7 @@ return JSON.stringify({
     };
   }),
   screen_reader_classes: captureAndLogError(() => {
-    return document.querySelectorAll('.sr-only, .visually-hidden', '.screen-reader-text', 'element-invisible').length > 0;
+    return document.querySelectorAll('.sr-only, .visually-hidden', '.screen-reader-text', '.element-invisible').length > 0;
   }),
   form_control_a11y_tree: captureAndLogError(() => {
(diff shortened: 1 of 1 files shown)


dist/a11y.js
```diff
@@ -215,7 +215,7 @@ return JSON.stringify({
     };
   }),
   screen_reader_classes: captureAndLogError(() => {
-    return document.querySelectorAll('.sr-only, .visually-hidden').length > 0;
+    return document.querySelectorAll('.sr-only, .visually-hidden', '.screen-reader-text', 'element-invisible').length > 0;
```
joedolson, line 218: I believe this should be `.element-invisible`, per https://www.drupal.org/node/2022859

mgifford: Oops.. Thanks @joedoleson..

mgifford: Oops.. Thanks @joedoleson..

tunetheweb: So can we update this?

```suggestion
    return document.querySelectorAll('.sr-only, .visually-hidden', '.screen-reader-text', '.element-invisible').length > 0;
```

tunetheweb (approved):

PR #143, opened 2024-09-13
Limit responsive image URLs

tunetheweb:
See: https://github.com/HTTPArchive/data-pipeline/issues/262#issuecomment-2348864332

Limit responsive image URLs to prevent massive data urls taking up lots of custom metric space.

---

**Test websites**:

- https://www.pronews.gr/

### DIFF
--- a/dist/responsive_images.js
+++ b/dist/responsive_images.js
@@ -305,10 +305,10 @@ function intrinsicOrExtrinsicSizing( computedStyles ) {
 function getImgData( img ) {
 
   const imgData = imgFeatures( img );
-  // Limit URLs to 200 chars so data URLs don't take up huge amounts of space.
+  // Limit data URLs so they don't take up huge amounts of space.
   imgData.url =
-    (img.currentSrc || img.src).length > 200 ?
-    (img.currentSrc || img.src).slice(0, 199) + '…' :
+    (img.currentSrc || img.src)?.startsWith('data:') && (img.currentSrc || img.src)?.length > 50 ?
+    (img.currentSrc || img.src)?.slice(0, 50) + '…' :
     (img.currentSrc || img.src);
   imgData.totalCandidates = totalNumberOfCandidates( img );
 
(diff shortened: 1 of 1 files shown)


dist/responsive_images.js
```diff
@@ -305,7 +305,11 @@ function intrinsicOrExtrinsicSizing( computedStyles ) {
 function getImgData( img ) {
 
   const imgData = imgFeatures( img );
-  imgData.url = img.currentSrc || img.src;
+  // Limit URLs to 200 chars so data URLs don't take up huge amounts of space.
```
tunetheweb, line 308: WDYT think @pmeenan @rviscomi @paulcalvano @max-ostapenko is 200 too much?

pmeenan: 200 should be fine but it might be better to put the elipsis in the middle and slice 100 off of each side so that the filename and extension can be kept (if there is one at the end).

tunetheweb: The other alternative is to record full URLs and just shortens data URLs. Could then be even stricter with first 10 chars or so.

@eeeps any thoughts here?

tunetheweb: I went with that approach instead.

tunetheweb:
@max-ostapenko any ideas why it didn't comment with the test site? It is in the artifact:

> Placeholder {artifact-url} replaced with a value: https://github.com/HTTPArchive/custom-metrics/actions/runs/10861083701/artifacts/1932874347.

max-ostapenko:
I added the artifact URL when the size is too big for a comment.

max-ostapenko:
I'm without a laptop to check archive content, but I see ZIP size is 40KB which definitely doesn't make sense for comment.

tunetheweb:
Ah I didn’t read the comment properly and thought that link was for the almanac test. Didn’t realise it was the other test. So all good.

max-ostapenko (approved):

PR #144, opened 2024-09-21
IAB Data Deletion Request Framework

max-ostapenko:
Newly introduced framework for programmatic [data deletion requests](https://iabtechlab.com/standards/data-deletion-request-framework/)

---

**Test websites**:

- https://data-deletion-request-framework.glitch.me/

### DIFF
--- a/dist/privacy.js
+++ b/dist/privacy.js
@@ -67,25 +67,25 @@ const isPresent = (response, endings) => response.ok && endings.some(ending => r
  * @returns {Promise<Object>} A promise that resolves to an object containing the parsed response data.
  */
 const parseDSRdelete = async (response) => {
-  let content;
-  try {
-    content = JSON.parse(await response.text());
-  } catch (error) {
-    console.error('Failed to parse JSON:', error);
-    content = null;
-  }
-
   let result = {
     present: isPresent(response, ['/dsrdelete.json']),
     redirected: response.redirected,
     status: response.status,
   };
 
-  if (result.present && content) {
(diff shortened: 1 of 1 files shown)


dist/privacy.js
```diff
@@ -25,7 +25,73 @@ function testPropertyStringInResponseBodies(pattern) {
...
+   * @param {function} parser - The function to parse the response.
+   * @returns {Promise<Object>} The parsed response or an error object.
+   */
+const fetchAndParse = async (url, parser) => {
+  const timeout = 5000;
+  /*
+  Google's sellers.json size is 120Mb as of May 2024 - too big for custom metrics.
+  It's available at realtimebidding.google.com/sellers.json, so not part of crawled pages list.
+  More details: https://support.google.com/authorizedbuyers/answer/9895942
+  */
```
tunetheweb, line 39: Is this needed? Or just copied from elsewhere?

tunetheweb (approved):
LGTM with one question.

dist/privacy.js
max-ostapenko: It's required here to clarify that the top ad tech player is not covered by this metric. And this bias should be considered during analysis.

max-ostapenko: We added this comment when implementing it in May 2024.

tunetheweb: Yes I understand that for ads.js. Is it the same for the re-use of this function in privacy.js? Is `/dsrdelete.json` also 120MB?

max-ostapenko: Sure, you're right. It's just copy-pasted comment.

Really need to check out the build tool asap.

PR #145, opened 2024-09-21
Limit data URLs in css-variables

tunetheweb:
As mentioned in https://github.com/HTTPArchive/data-pipeline/issues/262#issuecomment-2362212263

---

**Test websites**:

- https://www.morgenpost.de/politik/article407241629/game-changer-fuer-die-ukraine-bahnt-sich-an-putin-droht.html"

### DIFF
--- a/dist/css-variables.js
+++ b/dist/css-variables.js
@@ -46,6 +46,9 @@ function walkRule(rule, ret) {
 
     for (let property of style) {
       let value = style.getPropertyValue(property);
+      if (value.startsWith("url(\"data:image/svg+xml") && value.length > 54) {
+        value = value.slice(0, 50) + '…' + "\");";
+      }
 
       let containsRef = value.indexOf("var(--") > -1;
       let setsVar = property.indexOf("--") === 0 && property.indexOf("--" + PREFIX) === -1;


max-ostapenko:
It seems a `summary` part of the object is a duplicate (same content, different structure) of data available in `parsed_css` metric.

I'm able to find the variables/selectors/values there:
```sql
SELECT
  url, css
FROM `httparchive.all.parsed_css`
WHERE date = "2024-09-01"
AND client = 'mobile'
AND NOT is_root_page
AND rank <= 10000
and page = 'https://www.morgenpost.de/politik/article407241629/game-changer-fuer-die-ukraine-bahnt-sich-an-putin-droht.html'
AND css LIKE "%--playkit-icon-quality-HD-active-url%"
```

I can't give a guarantee, but it should be possible to reuse it for various metrics calculation.

The other part of the object - `custom_metrics.css-variables.computed` seems more unique.

So a quick optimization could be something as follows:
1. double-check if anybody is using `custom_metrics.css-variables.summary` ([last 90 days of logs](https://cloudlogging.app.goo.gl/ChbmdnXRPbNaR3fE6) don't show any queries),
2. drop it in favor of `parsed_css`,
3. BTW URLs are not shortened in `parsed_css` either. Apply this change [there](https://github.com/max-ostapenko/custom-metrics/blob/1f88d1f57011561b0a6cfe3cb22b93776b884b75/dist/parsed_css.js#L268) instead.

tunetheweb:
@rviscomi do you know the difference between these two custom metrics and whether we need both?

rviscomi:
[css-variables](https://github.com/HTTPArchive/custom-metrics/blob/main/dist/css-variables.js) gives us runtime data about custom property usage

[parsed_css](https://github.com/HTTPArchive/custom-metrics/blob/main/dist/parsed_css.js) gives us the raw stylesheet AST

They're similar in that they're both used for Web Almanac CSS analysis, but both are needed

tunetheweb:
OK, then I'm less worried about triming `parsed_css` data URLs since it's in it's own table and massive anyway. Plus I'm not 100% confident of what change to make and where.

I do think there's an argument to move both `css_variables` and `parsed_css` into the one table (and thus reduce the size of `custom_metrics` column in the main `all.pages` table some more) but let's leave that until the after the Web Almanac is finished.

So in meantime I think this is good to merge?

max-ostapenko:
We definitely need to split `css_variables` due to the size, [added to the list](https://docs.google.com/spreadsheets/d/14200qEV9QcXifjDbt-6uSUU7O7qmfYKevUN32XDWHCM/edit?gid=0#gid=0).

And yeah, good to merge.

max-ostapenko (approved):

PR #148, opened 2024-09-25
Resolved issue with is_lcp_statically_discoverable

kevinfarrugia:
The raw HTML (from response_body) uses relative URLs for images, for example "/static/img/logo.png", while calling .src on a document with a location, gives you absolute URLs, such as https://happymod.com/static/img/logo.png.

Consequently, when checking for the LCP element, we need to wrap the src with new URL to get an absolute URL. 

---
**Test websites**:

- https://happymod.com/
- https://httparchive.org/

### DIFF
--- a/dist/performance.js
+++ b/dist/performance.js
@@ -306,7 +306,7 @@ function getRawLcpElement(rawDoc, lcpUrl) {
             src = i.getAttribute('href');
         }
 
-        return src == lcpUrl;
+        return new URL(src, location.href).href == lcpUrl;
     });
 
     // lcp_elem_stats will match the img, not the source.


dist/performance.js
```diff
@@ -306,7 +306,7 @@ function getRawLcpElement(rawDoc, lcpUrl) {
             src = i.getAttribute('href');
         }
 
-        return src == lcpUrl;
+        return new URL(src, location.href).href == lcpUrl;
```
rviscomi, line 309: Can you also wrap the srcset check above in new URL?

> src = splitSrcSet(i.srcset).find(src => src == lcpUrl);

tunetheweb:
@kevinfarrugia , @rviscomi is this good to merge?

rviscomi:
The change LGTM, I'm just not sure why the GH action isn't testing the URLs in the description

rviscomi (approved):

max-ostapenko:
> The change LGTM, I'm just not sure why the GH action isn't testing the URLs in the description

`Test websites:` must be bold.

PR #149, opened 2024-10-07
Test remove summary metrics

max-ostapenko:
Metrics also available in summary object. Testing the origin of this data.

**Test websites**:

- https://example.com/


max-ostapenko:
@pmeenan I see duplicated metrics with different names in the test run.
Example link: http://webpagetest.httparchive.org/results.php?test=241007_TG_2


Screenshot 2024-10-07 at 22 02 41

max-ostapenko:
Ah, I see that these names are defined in the GitHub action code.

PR #157, opened 2025-01-27
Detect WordPress theme slugs

swissspidy:
There is a [new change](https://core.trac.wordpress.org/changeset/59698) in the upcoming WordPress 6.8 release that exposes the slugs of the currently active theme(s) via a class on the `<body>`.

This makes it easy to track individual theme adoption.

---

**Test websites**:

- https://ma.tt/

### DIFF
--- a/dist/cms.js
+++ b/dist/cms.js
@@ -13,26 +13,24 @@ function usesBlockTheme() {
  */
 function getWordPressTheme() {
   const theme = {
-    theme: 'unknown',
-    parent_theme: '',
+    theme: '',
+    child_theme: '',
   };
   try {
     const bodyClass = document.body.classList;
 
     const parentTheme = Array.from( bodyClass ).find( c => c.startsWith( 'wp-theme-' ) );
+
+    if ( parentTheme ) {
+      theme.theme = parentTheme.replace( 'wp-theme-', '' );
+    }
+
     const childTheme = Array.from( bodyClass ).find( c => c.startsWith( 'wp-child-theme-' ) );
 
     if ( childTheme ) {
-      theme.theme = childTheme;
-
-      // If there is a child theme, there should always be a parent theme.
-      if ( parentTheme ) {
-        theme.parent_theme = parentTheme;
-      }
-    } else if ( parentTheme ) {
-      // There is no child theme, only a parent theme.
-      theme.theme = parentTheme;
+      theme.child_theme = childTheme.replace( 'wp-child-theme-', '' );
     }
+
   } catch ( e ) {}
   return theme;
 }
--- a/dist/cms.js
+++ b/dist/cms.js
@@ -13,8 +13,8 @@ function usesBlockTheme() {
  */
 function getWordPressTheme() {
   const theme = {
-    theme: '',
-    child_theme: '',
+    theme: null,
+    child_theme: null,
   };
   try {
     const bodyClass = document.body.classList;
@@ -23,6 +23,7 @@ function getWordPressTheme() {
 
     if ( parentTheme ) {
       theme.theme = parentTheme.replace( 'wp-theme-', '' );
+      theme.child_theme = '';
     }
 
     const childTheme = Array.from( bodyClass ).find( c => c.startsWith( 'wp-child-theme-' ) );
--- a/dist/cms.js
+++ b/dist/cms.js
@@ -4,6 +4,39 @@ function usesBlockTheme() {
   return !!document.querySelector('div.wp-site-blocks');
 }
 
+/**
+ * Detects the WordPress parent and child theme slugs.
+ *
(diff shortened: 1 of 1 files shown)


dist/cms.js
```diff
@@ -4,6 +4,39 @@ function usesBlockTheme() {
...
+ *
+ * See https://core.trac.wordpress.org/changeset/59698.
+ *
+ * @returns {object} Object with fields `theme` and `parent_theme`.
+ */
+function getWordPressTheme() {
+  const theme = {
+    theme: 'unknown',
+    parent_theme: '',
+  };
```
felixarntz, line 18: From a tracking perspective, I think it might be better to go with `theme` and `child_theme` (or explicitly `parent_theme` and `child_theme`), in line with how these classes are added.

The conditional approach you're currently using is most common in WordPress terminology, but with that approach any theme that is not made as a child theme could appear under either `theme` or `parent_theme`, so it makes tracking more difficult than needed.

So I think we should always put the parent theme and child theme under the same respective property.

dist/cms.js
```diff
@@ -4,6 +4,39 @@ function usesBlockTheme() {
...
+      theme.theme = childTheme;
+
+      // If there is a child theme, there should always be a parent theme.
+      if ( parentTheme ) {
+        theme.parent_theme = parentTheme;
+      }
+    } else if ( parentTheme ) {
+      // There is no child theme, only a parent theme.
+      theme.theme = parentTheme;
+    }
```
felixarntz, line 35: Based on my comment above, a nice side effect is that this logic could be simplified because we would align with how WP Core applies these classes (`wp-theme-` is always parent theme, `wp-child-theme-` is always child theme).

felixarntz (commented):
@swissspidy Thanks for the PR! Looks good overall, though I think tracking this slightly differently would be more efficient for queries.

dist/cms.js
```diff
@@ -4,6 +4,39 @@ function usesBlockTheme() {
...
+ * Detects the WordPress parent and child theme slugs.
+ *
+ * See https://core.trac.wordpress.org/changeset/59698.
+ *
+ * @returns {object} Object with fields `theme` and `parent_theme`.
+ */
+function getWordPressTheme() {
+  const theme = {
+    theme: 'unknown',
+    parent_theme: '',
```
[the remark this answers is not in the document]
swissspidy: yeah just an oversight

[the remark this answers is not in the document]
swissspidy: note that `unknown` could be the actual name of a theme, so I'll go with empty string I think

dist/cms.js
swissspidy: > The conditional approach you're currently using is most common in WordPress terminology, but with that approach any theme that is not made as a child theme could appear under either `theme` or `parent_theme`

True. I intended for this to make it easier to track a specific theme being used (simply always check `theme.theme`, but I didn't fully think it through.

But I guess if you want to know how often a specific theme is actually used (without child themes), then one could do `theme === xyz && child_theme === ''``

dist/cms.js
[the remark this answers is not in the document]
felixarntz: How about `null`? That should work well and is in line with regular "not set" values in BigQuery.

dist/cms.js
```diff
@@ -4,6 +4,37 @@ function usesBlockTheme() {
...
+ * Detects the WordPress parent and child theme slugs.
+ *
+ * See https://core.trac.wordpress.org/changeset/59698.
+ *
+ * @returns {object} Object with fields `theme` and `parent_theme`.
+ */
+function getWordPressTheme() {
+  const theme = {
+    theme: '',
+    child_theme: '',
```
felixarntz, line 17: This works, though related to https://github.com/HTTPArchive/custom-metrics/pull/157#discussion_r1931143821 it might still be good to differentiate between "unknown" and "none". I'd say if `wp-theme-...` is present and `wp-child-theme-...` is not present, we should set `child_theme` to `''` (none).

But if neither of the two are present, we for sure don't know (because every site _must_ have at least a theme). So then using another value for both (e.g. `null`) would be good.

swissspidy: So null if both are absent? Works for me.

felixarntz: Exactly.

felixarntz: Maybe also `theme: null` if for some reason a `wp-child-theme-...` class is present, but `wp-theme-...` isn't. That would be invalid and is probably an edge-case, so might also fine to ignore it.

felixarntz (approved):
@swissspidy LGTM!

tunetheweb (approved):
LGTM. I presume it's not possible to have multiple themes or child themes?

felixarntz:
> LGTM. I presume it's not possible to have multiple themes or child themes?

Indeed, that's not possible. 👍

PR #158, opened 2025-02-04
Add support for FedCM, Passkey, and WebAuthn well-known files

tsunoyu:
## Description
This PR adds support for tracking and analyzing the following well-known files:
1. **FedCM**: `/.well-known/web-identity`
2. **Passkey**: `/.well-known/passkey-endpoints`
3. **WebAuthn**: `/.well-known/webauthn`

These files are critical for enabling **Federated Credential Management (FedCM)**, **passkey upgrades**, and **WebAuthn-related origin requests**. The changes ensure that these files are fetched, parsed, and included in the HTTP Archive's custom metrics.

## Changes
1. Added a new `parseResponse` call for `/.well-known/web-identity` to track FedCM configurations.
2. Added a new `parseResponse` call for `/.well-known/passkey-endpoints` to track passkey enrollment and management URLs.
3. Added a new `parseResponse` call for `/.well-known/webauthn` to track WebAuthn-related origin requests.

### FedCM (`/.well-known/web-identity`)
```json
{
  "/.well-known/web-identity": {
    "found": true,
    "data": {
      "provider_urls": ["https://accounts.idp.example/config.json"],
      "accounts_endpoint": "https://accounts.idp.example/accounts",
      "login_url": "https://accounts.idp.example/login"
    }
  }
}

{
  "/.well-known/passkey-endpoints": {
    "found": true,
    "data": {
      "enroll": "https://passkeys-demo.appspot.com/home",
      "manage": "https://passkeys-demo.appspot.com/home"
    }
  }
}

{
  "/.well-known/webauthn": {
    "found": true,
    "data": {
      "origins": ["https://example.com"]
    }
  }
}
```
---

**Test websites**:

- https://www.ebay.com/
- https://www.amazon.co.uk/
- https://tv.apple.com/
- https://ror-1.glitch.me/
- https://www.google.com/


.github/pull_request_template.md
```diff
@@ -7,4 +7,6 @@ Description of the changes...
 <!-- List the pages that should be automatically tested as part of your custom metric changes. -->
 **Test websites**:
 
-- https://example.com/
+- https://www.ebay.com/
+- https://www.amazon.co.uk/
+- https://tv.apple.com/
```
tunetheweb, line 12: Please remove from the template.

dist/well-known.js
```diff
@@ -115,6 +115,34 @@ return Promise.all([
       return data;
     });
   }),
+  // FedCM
+  parseResponse('/.well-known/web-identity', r => {
```
tunetheweb, line 119: Linting error.

```suggestion
  parseResponse('/.well-known/web-identity', r => {
```

dist/well-known.js
```diff
@@ -115,6 +115,34 @@ return Promise.all([
       return data;
     });
   }),
+  // FedCM
+  parseResponse('/.well-known/web-identity', r => { 
+    return r.json().then(data => {
```
tunetheweb, line 120: Should there be error handling here? There are quite a few errors in the `tv.apple.com` example run for example.

tunetheweb (changes requested):

.github/pull_request_template.md
tunetheweb:
```suggestion
```

dist/well-known.js
tsunoyu: Thank you for pointing it out. I will update the code with additional error handling.

tsunoyu: ##  Case with Apple Domain , Update on Error Handling  

During testing with `tv.apple.com`, we encountered a case where `.well-known` endpoints returned **200 OK** responses but contained **invalid JSON** instead of expected structured data.  

### Issue Explanation  
Ideally, if an endpoint is not available, it should return **404 Not Found** or **403 Forbidden** along with an appropriate HTML message. However, Apple’s servers seem to **return a 200 OK response** while serving **their homepage HTML** instead of a valid JSON response.  

Rather than implementing a **301 redirect**, they send back a **200 response with their homepage content**, which violates expected standards and complicates JSON validation.  

### Implemented Error Handling  
- Added a `"valid"` status field to verify whether the response is **both found and correctly formatted**.  
- Improved JSON parsing to ensure that even if an endpoint exists, we can determine whether the response meets expected requirements.  


#### **Test Results**:
```json
"/.well-known/web-identity": {
  "found": true,
  "data": {
    "error": "Failed to parse JSON: Unexpected token '<', \"<!DOCTYPE \"... is not valid JSON",
    "valid": false
  }
},
"/.well-known/passkey-endpoints": {
  "found": true,
  "data": {
    "error": "Failed to parse JSON: Unexpected token '<', \"<!DOCTYPE \"... is not valid JSON",
    "valid": false
  }
},
"/.well-known/webauthn": {
  "found": true,
  "data": {
    "error": "Failed to parse JSON: Unexpected token '<', \"<!DOCTYPE \"... is not valid JSON",
    "valid": false
  }
}
```

@tunetheweb Would it be possible to get your review?

tunetheweb: Do we need the `valid` attribute? Isn't the presence of the `error` enough to show it wasn't valid?

In which case your original implementation may have been fine as it was if those are expected failures (based on their invalid response)?

tsunoyu: @tunetheweb Thanks for the quick review!
The `valid` attribute does seem redundant given the presence of the `error` message.  I agree that simplifying the output is better approach here.

I'll revert to commit d6c5dbe7b20ba01656cccc98af98f96659e14628, which handles the invalid JSON responses correctly without the extra `valid` field.

tunetheweb: Just checking you're still working on this? As still see `valid` lines on 126, 132, 142, 148, 157, and 163.

dist/well-known.js
```diff
@@ -115,6 +116,54 @@ return Promise.all([
...
+        };
+    });
+  }),
+  // Passkey
+  parseResponse('/.well-known/passkey-endpoints', r => {
+    return r.json().then(data => {
+        let result = {
+            enroll: data.enroll || null,
+            manage: data.manage || null,
+            valid: typeof data.enroll === 'string' && typeof data.manage === 'string'
```
max-ostapenko, line 142: Indeed seems like `valid` duplicates the other two.
```suggestion
            enroll: data.enroll || null,
            manage: data.manage || null
```

dist/well-known.js
```diff
@@ -115,6 +116,54 @@ return Promise.all([
...
     });
   }),
+  // FedCM
+  parseResponse('/.well-known/web-identity', r => {
+    return r.json().then(data => {
+        let result = {
+            provider_urls: Array.isArray(data.provider_urls) && data.provider_urls.length > 0 ? data.provider_urls : [],
+            accounts_endpoint: data.accounts_endpoint || null,
+            login_url: data.login_url || null,
+            valid: Array.isArray(data.provider_urls) && data.provider_urls.length > 0
```
max-ostapenko, line 126: `valid` is based on `provided_urls` so don't need to log it and can analyze it when querying the data:
```suggestion
            provider_urls: Array.isArray(data.provider_urls) && data.provider_urls.length > 0 ? data.provider_urls : [],
            accounts_endpoint: data.accounts_endpoint || null,
            login_url: data.login_url || null
```

dist/well-known.js
max-ostapenko: @tsunoyu You're right, it's probably 3rd most popular scenario - getting a 200 with HTML saying 'Not found'.
And I think the error message is too big, for it's value, especially when it's gonna be all the same when we are trying to parse HTML as JSON.


Maybe we can go the other way round: isn't the presence of `provider_urls` or `login_url` not enough to identify the valid ones?

max-ostapenko (changes requested):

dist/well-known.js
```diff
@@ -115,6 +116,54 @@ return Promise.all([
...
+            origins: Array.isArray(data.origins) && data.origins.length > 0 ? data.origins : [],
+            valid: Array.isArray(data.origins) && data.origins.length > 0
+        };
+        return result;
+    }).catch(error => {
+        return {
+            error: `Failed to parse JSON: ${error.message}`,
+            valid: false
+        };
+    });
```
max-ostapenko, line 165: Again it's gonna be much more invalid ones than valid, so let's avoid storing error information.
```suggestion
    });
```

max-ostapenko (changes requested):

dist/well-known.js
```diff
@@ -115,6 +116,48 @@ return Promise.all([
...
+            accounts_endpoint: data.accounts_endpoint || null,
+            login_url: data.login_url || null
+        };
+        return result;
+    }).catch(error => {
+        return {
+            error: `Failed to parse JSON: ${error.message}`
+        };
+    });
+  }),
```
tsunoyu, line 133: @max-ostapenko @tunetheweb Thank you for your previous feedback. I was trying to reflect your feedbacks and made the update - not producing an error for invalid "JSON" any more, instead we are responding with default results.
Does this update align better? If this update looks good for you, I will update for the other two parts ('/.well-known/passkey-endpoints' and '/.well-known/webauthn') as well.  

```suggestion
  parseResponse('/.well-known/web-identity', r => {
    return r.text().then(text => {
        let result = {
            provider_urls: [],
            accounts_endpoint: null,
            login_url: null
        };
        try {
            let data = JSON.parse(text);
            result.provider_urls = Array.isArray(data.provider_urls) && data.provider_urls.length > 0 ? data.provider_urls : [];
            result.accounts_endpoint = data.accounts_endpoint || null;
            result.login_url = data.login_url || null;
        } catch (e) {
            // Failed to parse JSON
        }
        return result;
    });
  })
```

max-ostapenko: Looks good.

Among the test websites only https://www.ebay.com/.well-known/passkey-endpoints gives a successful result.
Do you know the examples for the other 2?

max-ostapenko:
Added a WebAuthn example (https://ror-1.glitch.me/) from https://web.dev/articles/webauthn-related-origin-requests

tsunoyu:
> Added a WebAuthn example (https://ror-1.glitch.me/) from https://web.dev/articles/webauthn-related-origin-requests
> 
> UPD. + https://www.google.com/.well-known/web-identity

max-ostapenko (approved):

tunetheweb (approved):

PR #160, opened 2025-03-05
Add eslint linting

tunetheweb:
Looks like this recently got enabled, but a lot of our custom metrics fail. And often for very good reasons.

Fixed them all up and think they look OK now, but there's a lot of change here so one to keep an eye on in next crawl.

---

**Test websites**:

- https://example.com/

### DIFF
--- a/dist/avg_dom_depth.js
+++ b/dist/avg_dom_depth.js
@@ -12,7 +12,8 @@ function avgDomDepth() {
 function numParents(elem) {
   var n = 0;
   if ( elem.parentNode ) {
-    while ( elem = elem.parentNode) {
+    /* eslint-disable-next-line no-cond-assign */
+    while (elem = elem.parentNode) {
       n++;
     }
   }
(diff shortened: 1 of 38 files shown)


dist/avg_dom_depth.js
```diff
@@ -12,7 +12,8 @@ function avgDomDepth() {
 function numParents(elem) {
   var n = 0;
   if ( elem.parentNode ) {
-    while ( elem = elem.parentNode) {
+    /* eslint-disable-next-line no-cond-assign */
+    while (elem = elem.parentNode) {
```
max-ostapenko, line 16:
```suggestion
    while ((elem = elem.parentNode)) {
```

max-ostapenko (approved):

dist/avg_dom_depth.js
tunetheweb: Can you explain why this would be needed out of interest?

max-ostapenko: Just an alternative approach that I found easier than adding ignore.
It hides assignment within parenthesis and linter is fine.

tunetheweb: Oh I missed that you'd removed the linter override too. Nice work around (if a little bit of a cheat!).

PR #161, opened 2025-03-05
No unnecessary escaping, but the necessary escaping yes

tomayac:


dist/fugu-apis.js
```diff
@@ -486,7 +486,7 @@ const patterns = {
         ((await navigator.serviceWorker?.ready) || self.registration))(),
     featureDetection: `(async () => 'serviceWorker' in navigator && 'pushManager' in (await navigator.serviceWorker?.ready || self.registration))()`,
     documentation: 'https://developer.mozilla.org/en-US/docs/Web/API/Push_API',
-    blinkFeatureID: 769,
+    blinkFeatureID: 371,
```
tunetheweb, line 489: Did this feature id change?

tomayac: Yepp. A better one was found.

tunetheweb:
Including these changes in #160

PR #165, opened 2025-06-13
Added exposed API key custom metric

vsdaan:
Added custom metric to check for exposed API keys in both inline scripts and external sources.

---

**Test websites**:

- https://example.com/
- https://daanvs.be/apitest
First test should return an empty array, the second test should flag gemini and openai.


max-ostapenko:
@vsdaan let me know when you're ready with it.

nrllh:
While I like the idea, I don’t think it’s a good idea to include this. This custom metric could allow attackers to easily find API keys and exploit them. Therefore, I recommend not merging this PR.

max-ostapenko:
Yes, this will definitely be a negative experience delivered by HTTP Archive.
And reaching out to the website owners privately is not the priority of Web Almanac team.

It may just bring attention of attackers to the dataset, which is not our goal.

PR #168, opened 2025-06-24
Update README.md

max-ostapenko:
Updated readme to represent an actual HTTP Archive workflow. Also webpagetest.org website has different UI.

- [x] Updated URLs in the "Manual testing using webpagetest.org website" section to use `webpagetest.httparchive.org` instead of `webpagetest.org`.

- [x] Updated the link to the WPT API wrapper in the "Automated WPT test runs" section to point to the correct repository at `HTTPArchive/WebPageTest.api-nodejs`.


pmeenan (approved):
LGTM, thanks.

PR #169, opened 2025-06-24
Added cloud genAI service detection

JonathanPLev:
Added a custom metric to detect which cloud genAI services a website accesses.

---
**Test websites**:

- https://example.com/
- https://chatgpt.com/
- https://www.yes4youth.co.za/4-business?hsLang=en

### DIFF
--- a/dist/cloud_genAI_services.js
+++ b/dist/cloud_genAI_services.js
@@ -1,0 +1,114 @@
+//[genAI-Service-Detection]
+// uncomment for webpagetest
+
+// not sure if the list is exhaustive
+console.log("Running genAI-Service-Detection metric");
+try{
+const patterns = {
+    // GenAI Services
+    'Azure AI Services': {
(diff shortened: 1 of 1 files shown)


JonathanPLev:
@max-ostapenko is adding my metrics to that repo the same process as for this one? I don't see information on how to add to that repo in that ReadME.

alternatively, would I be able to adjust my detection or add some sort of other analysis that would qualify the detection for this repo?

Thank you!

max-ostapenko:
@JonathanPLev here is a recent example of technology added: https://github.com/HTTPArchive/wappalyzer/pull/136

I also notice that you use API endpoint URLs - these are usually used on the backend. There is also an argument against detecting such usage scenarios: https://github.com/HTTPArchive/custom-metrics/pull/165#issuecomment-3007011195

There usually should be some JS library, or JS variable that help detect technology on the webpage. Please see the patterns supported for detection: https://github.com/HTTPArchive/wappalyzer?tab=readme-ov-file#patterns-optional

PR #170, opened 2025-06-29
Add HTTP Header speculation rules

### LINKED ISSUE #111: Improve coverage of Speculation Rules custom metrics
In [performance.js](https://github.com/HTTPArchive/custom-metrics/blob/1453b36b96cbb1773aa362d86d90c4d59b28964d/dist/performance.js#L327-L335) we're capturing the speculation rules themselves. This tells us how the links should be prefetched/prerendered, but not which ones actually are.

Add new metadata to the custom metric that captures which requests are actually speculatively loaded. We can look for the [`Sec-Purpose`](https://developer.chrome.com/docs/web-platform/prerender-pages#detecting_and_disabling_prerender_server-side) header and log which requests are prefetched or prerendered.

The custom metric also currently only looks for script tags with the Speculation Rules JSON. But it's also possible to configure the rules via the `Speculation-Rules` header. The custom metric should check there as well as the script tags.

tunetheweb:
Resolves #111

Fetches HTTP header Speculation Rules and stores them in the `performance` custom metric.

**Test websites**:

- https://example.com/
- https://blkandbold.com/
- https://www.hillhousehome.com/
- https://wordpress.com/

### DIFF
--- a/dist/performance.js
+++ b/dist/performance.js
@@ -3,6 +3,12 @@
 const response_bodies = $WPT_BODIES;
 const script_response_bodies = $WPT_BODIES.filter(body => body.type === 'Script');
 
+function fetchWithTimeout(url) {
+  var controller = new AbortController();
+  setTimeout(() => {controller.abort()}, 5000);
+  return fetch(url, {signal: controller.signal});
+}
+
 function getRawHtmlDocument() {
     let rawHtml;
     if (response_bodies.length > 0) {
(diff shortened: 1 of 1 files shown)


dist/performance.js
```diff
@@ -324,17 +330,53 @@ function getLcpResponseObject(lcpUrl) {
...
+  const htmlRules = Array.from(document.querySelectorAll('script[type=speculationrules]')).map(script => {
+      try {
+          return JSON.parse(script.innerHTML);
+      } catch (error) {
+          return null;
+      }
+  });
+
+  // Get rules from the HTTP requests
+  const documentRequests = [$WPT_REQUESTS.find( req => req.url === document.location.href)] || [];
```
tunetheweb, line 348: @pmeenan is this the best way of getting the document request? Just `find`ing the first `$WPT_REQUEST` that matches the document.location.href?

One thing that could break is if the page updates the href after fetching (e.g. in SPAs), so ideally there'd be a better way.

tunetheweb: Went with this instead:

```js
const documentRequest = $WPT_REQUESTS.find(
   req => req.url === performance.getEntriesByType('navigation')[0].name
);
```

dist/performance.js
```diff
@@ -324,17 +330,52 @@ function getLcpResponseObject(lcpUrl) {
...
-            return JSON.parse(script.innerHTML);
-        } catch (error) {
-            return null;
+          let url = decodeURI(speculationRuleLocation).slice(1, -1);
+          if (url.startsWith('/')) {
+            url = document.location.origin + url;
+          } else if (!url.startsWith('http')) {
+            url = document.location.href + url;
+          }
+          const response = await fetchWithTimeout(url);
```
pmeenan, line 365: Worth checking if there is already a request in the $WPT_REQUESTS for the speculation rules that you can parse directly without fetching?

Better-yet, if there is a way to identify a speculation rules fetch directly (dest?), default to using that if there is one before doing any of this processing.

pmeenan (approved):
LGTM though wondering if it might be possible to pull the speculation rules fetch itself out of the $WPT_REQUESTS data.

pmeenan (approved):
LGTM.  If the `Sec-Fetch`Dest` request header is available, it might be more accurate to use that but I'm not sure if the header is available at custom-metrics time.

dist/performance.js
tunetheweb: Cool. Went with the below to check `content-type` and `sec-fetch-dest`. A much simpler implementation that what I had before.

```js
  // Get rules from Speculation-Rules HTTP responses
  // There is an assumption this is actually used on the page but by checking both the `content-type`
  // and the `sec-fetch-dest`, that should be the case.
  const httpRules = Array.from(
    response_bodies
    .filter(req => getParameterCaseInsensitive(req.response_headers, 'content-type') === 'application/speculationrules+json')
    .filter(req => getParameterCaseInsensitive(req.request_headers, 'sec-fetch-dest') === 'speculationrules')
    .map(req => {
      try {
        return {url: req.url, rule: JSON.parse(req.response_body)};
      } catch(error) {
        return {url: req.url, errorName: error.name, errorMessage: error.message};
      }
    })
  );
  ```

PR #174, opened 2025-09-16
Add error checking on performance custom metric

tunetheweb:
Speculation Rules custom metric is causing the performance custom metric to fail with error sometimes with:

```json
{"errorMessage":"Cannot convert undefined or null to object","errorName":"TypeError"}
```

---

**Test websites**:

- https://example.com/
- https://www.swp.de/lokales/horb/tanklaster-mit-loesungsmittel-brennt-400-einsatzkraefte-17-stunden-so-lief-der-grosseinsatz-auf-der-a-81-78186666.html

### DIFF
--- a/dist/performance.js
+++ b/dist/performance.js
@@ -325,6 +325,7 @@ function getLcpResponseObject(lcpUrl) {
 }
 
 function getParameterCaseInsensitive(object, key) {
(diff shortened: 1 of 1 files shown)


pmeenan (approved):
LGTM

PR #175, opened 2025-09-16
Add error checking to origin-trials custom metric

tunetheweb:
Same fix as for #174 for origin-trial custom metric

---

**Test websites**:

- https://example.com/


pmeenan (approved):
LGTM

PR #176, opened 2025-09-18
Update agent-card.json for A2A Protocol

tsunoyu:
## Description

This PR adds support for tracking and analyzing a critical file related to AI Agent discovery and communication:

1.  **Agent Card**: `/.well-known/agent-card.json`

This file is central to the **[Agent2Agent (A2A) Protocol](https://a2a-protocol.org/latest/)** and serves as the agent's digital business card. It's used by clients to discover the agent's identity, capabilities, service endpoints, and security requirements. The changes ensure this file is fetched, parsed, and included in the HTTP Archive's custom metrics for analysis of emerging AI agent standards adoption.

## Changes

1.  Added a new `parseResponse` call for `/.well-known/agent-card.json` to track its presence, structure, and key metadata (e.g., agent name, version, and declared skills).

### Agent Card (`/.well-known/agent-card.json`)
```json
{
  "/.well-known/agent-card.json": {
    "found": true,
    "data": {
      "url": "https://agent.example.com/a2a",
      "preferredTransport": "http",
      "protocolVersion": "0.3.0",
      "name": "Example Service Agent",
      "description": "An agent that provides example services via A2A."
    }
  }
}
```

---

**Test websites**:

- https://www.google.com/
- https://www.mozilla.org/
- https://www.microsoft.com/
- https://www.cloudflare.com/
- https://www.github.com/

### DIFF
--- a/dist/well-known.js
+++ b/dist/well-known.js
@@ -311,6 +311,29 @@ return Promise.all([
       url: r.url
     });
   }),
+  // A2A Agent Card (required + strongly recommended fields)
+  parseResponse('/.well-known/agent-card.json', r => {
+    return r.text().then(text => {
+      let result = {
+        url: null,
+        preferredTransport: null,
+        protocolVersion: null,
+        name: null,
+        description: null
+      };
+      try {
+        let data = JSON.parse(text);
+        result.url = data.url || null;
+        result.preferredTransport = data.preferredTransport || null;
+        result.protocolVersion = data.protocolVersion || null;
+        result.name = data.name || null;
+        result.description = data.description || null;
+      } catch (e) {
+        // Failed to parse JSON
+      }
+      return result;
+    });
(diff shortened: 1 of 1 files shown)


tunetheweb:
None of the example sites are detecting this. Is this cause they don't publish an agent-card.json or cause the check is wrong? Do we have an example of a site that DOES publish this to test the detection works?

tunetheweb:
Thanks. You need to do a dummy commit (or a force push?) to retrigger the CI. Unfortunately rerunning doesn’t work as it picks up the original list.

tunetheweb:
Urgh. Looks like because the site doesn’t return a 200 response we don’t bother running custom metrics on it 😔

tsunoyu:
I will try to find a different one with status 200. Let me check further. Thank you for your support.

PR #197, opened 2026-08-18
feat: implement JSDoc-based documentation validation and automated MDX generation workflow

max-ostapenko:
This pull request introduces automated workflows and supporting scripts for validating and syncing JSDoc-based documentation with code. The primary focus is on ensuring documentation parity and providing a mechanism to generate and synchronize documentation with the `har.fyi` repository.

Example of the validated and generated docs: https://github.com/HTTPArchive/har.fyi/pull/167

**GitHub Actions Workflows:**

* Added a new workflow `.github/workflows/docs.yml` to validate JSDoc comments against code and automatically sync generated documentation to the `har.fyi` repository via a pull request. This workflow runs on pushes, pull requests, and can be triggered manually.
* Updated `.github/workflows/linter.yml` to include a new `validate-docs` job that runs the documentation validation as part of the CI process.

**Documentation Scripts and Tooling:**

* Added new npm scripts `validate:docs` and `generate:docs` in `package.json` to run documentation validation and generation tools.


.github/workflows/docs.yml
```diff
@@ -0,0 +1,93 @@
...
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20
+          cache: 'npm'
+
+      - name: Install Dependencies
+        run: npm ci
+
+      - name: Validate JSDoc vs Code
+        run: npm run validate:docs
```
tunetheweb, line 39: What does this validate step do and will it fail if you make a change (and so they are out of sync for now until below sync runs)?

.github/workflows/linter.yml
```diff
@@ -68,3 +68,23 @@ jobs:
...
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20
+          cache: "npm"
+
+      - name: Install Dependencies
+        run: npm ci
+
+      - name: Validate JSDoc vs Code
+        run: npm run validate:docs
```
tunetheweb, line 89: Same question here.

dist/ads.js
```diff
@@ -27,7 +32,35 @@ const fetchAndParse = async (url, parser) => {
   }
 };
 
-// https://iabtechlab.com/wp-content/uploads/2022/04/Ads.txt-1.1.pdf
```
tunetheweb, line 30: We've lost this comment. Is it important to keep?

dist/ads.js
```diff
@@ -100,7 +133,35 @@ const parseAdsTxt = async (response) => {
 }
 
 
-// https://iabtechlab.com/wp-content/uploads/2019/07/Sellers.json_Final.pdf
```
tunetheweb, line 103: Same comment.

dist/privacy.js
```diff
@@ -83,26 +111,24 @@ let sync_metrics = {
...
+   * @typedef {Object} IABTCFv1
+   * @property {boolean} present - Whether the `__cmp` API function is present on the window object.
+   * @property {Object} [data] - TCF v1 vendor consents data returned by `getVendorConsents`. See [VendorConsents](https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/blob/master/CMP%20JS%20API%20v1.1%20Final.md#vendorconsents-).
+   * @property {boolean} [compliant_setup] - Verifies whether the TCF v1 CMP setup is compliant with IAB standards.
    */
   iab_tcf_v1: (() => {
     let consentData = {
       present: typeof window.__cmp == 'function',
     };
-    // description of `__cmp`: https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/blob/master/CMP%20JS%20API%20v1.1%20Final.md#what-api-will-need-to-be-provided-by-the-cmp-
```
tunetheweb, line 92: Again losing lots of comments from this file, not all of which are copied to JSDocs. Is that intentional? Why?

tunetheweb (commented):
Looks mostly OK to me but have some comments.

My two main concerns would be:
1) Will this make it less likely for contributors to want to edit custom metrics? Probably not?
2) You're planning on migrating har.fyi to httparchive.org so if this the right time to introduce this?

dist/ads.js
max-ostapenko: I updated it to Github spec link, and moved to JSDoc - see `@property {AdsTxtData}`

.github/workflows/docs.yml
max-ostapenko: Validate step checks the matching of docs to code in both directions. If anything is missing - fails. It even does a small effort to check the types correctness.

The out-of-the-box validation solution can be to migrate to TypeScript and add compilation stage, but I decided against it because of potentially unfamiliar syntax to contributors.

Action doesn't fail on this branch so the JSDocs+code are in sync.

Upon push to main it will push the JSDocs into har.fyi markdown. So the assumption is JSdocs will replace editing har.fyi.

.github/workflows/linter.yml
max-ostapenko: Should we deduplicate it, or wdym?

max-ostapenko:
> 1. Will this make it less likely for contributors to want to edit custom metrics? Probably not?

That's why I decided to keep the code as-is, just require comments.
We can add a quick guide on JSDoc if there will be questions. But I hope by expanding it to other metrics will give plenty of examples.

> 2. You're planning on migrating har.fyi to httparchive.org so if this the right time to introduce this?

Switching to another destination repo is not a big effort.

dist/ads.js
max-ostapenko: moved to `@property {SellersJsonData} sellers`

dist/privacy.js
max-ostapenko: A few links indeed got lost in the rewrite. Fixed.

.github/workflows/docs.yml
tunetheweb: Sorry I'm still not getting it.

So I make a change that changes the JS Doc.
I open the PR.
It presumably checks my changes against product har.fyi and fails.
But you're saying "Action doesn't fail on this branch"? So what's the point of doing this check?

Assuming I get past that, I get my change reviewed and merged.
It now runs against `main`
It presumably checks my changes against product har.fyi and fails.
Does that fail the action then? If so isn't that bad? If not, then what's the point?

Assuming I get past that, it then runs the sync and brings it back inline after that PR is merged.

.github/workflows/linter.yml
tunetheweb: I'm questioning if we're going to get failures in CI every time we make a change that includes a JS Doc change since we're out of sync with har.fyi until the change is merged.

.github/workflows/docs.yml
max-ostapenko: har.fyi is strictly downstream storage for docs. There is no check between this repo and har.fyi

`validate:docs` checks JSDoc and the code itself in custom metrics files - the expectation is that when the code changes there will be a JSDoc describing it.
If this is satisfied, then `docs.yml` generates new MD and pushes new PR into har.fyi.

.github/workflows/linter.yml
max-ostapenko: No, none of CI's here depends on har.fyi.

.github/workflows/docs.yml
tunetheweb: Ahh ok. Got it now.

tunetheweb (approved):

PR #201, opened 2026-09-22
well-known: add /.well-known/ai-catalog.json (Agentic Resource Discovery)

flober81:
Adds `/.well-known/ai-catalog.json` to the well-known custom metric.

**Why:** Lighthouse ships an `agentic-browsing` category in its default config (`core/config/default-config.js`). Its ARD gatherer (`core/gather/gatherers/agentic/ard.js`) resolves the catalog via robots.txt `agentmap` → HTML `<link rel="ai-catalog">` → HTTP Link header → **fallback `/.well-known/ai-catalog.json`**, and `agentic/ard-schema` validates it. The audit also runs in PageSpeed Insights. So this path now has a checker in a widely used tool, but HTTP Archive does not record it yet.

**What is recorded:** same shape as the other JSON well-known entries — `found` from the wrapper, plus `spec_version`, `has_host`, `entries_count` and the distinct `entry_types` (capped at 20). Invalid JSON or an HTML catch-all page yields the default values, no error. Kept to a single path on purpose (`llms.txt` already has its own metric, `agent-card.json` is #176).

**Baseline for reference:** in a monthly panel of ~850 hosts (AI Discovery Radar, DOI 10.5281/zenodo.22178282) the route is at 0.00 % as of 2026-09; HTTP Archive would be the first place to see whether the Lighthouse check moves it.

---
**Test websites**:

- https://btlabs.dev/ (serves the file with 200, `specVersion` 1.0, 3 entries)
- https://example.com/ (negative control, expected `found: false`)


flober81:
The Tests workflow fails at checkout, not on the change: since the actions/checkout v7 bump in #199, `pull_request_target` refuses fork checkouts unless `allow-unsafe-pr-checkout` is set. This is the first fork PR since then. Happy to adjust anything on the metric itself.

max-ostapenko:
@flober81 we'll unblock this soon with #202

max-ostapenko:
@flober81 can we use destination from `link rel` instead of default when available?
Interesting if some platforms may prefer custom path.

flober81:
Done in 28fba51: when the document has `<link rel="ai-catalog">`, its target is fetched instead of the default path (same order as Lighthouse's ARD gatherer). The result is still keyed under `/.well-known/ai-catalog.json` so the BigQuery key stays stable; `catalog_url` and `source` (`link` / `well-known`) record what was actually fetched, which answers the "custom path" question directly. Cross-origin link targets are subject to CORS like the other fetches here.

max-ostapenko:
@flober81 let's save a few bytes in the crawl results:

1. can keep a shorter version when nothing is found:
```json
    "/.well-known/ai-catalog.json": {
      "found": false
    },
```

2. and deduplicate `"catalog_url": "/.well-known/ai-catalog.json",` - add it only when `link.origin !== location.origin || link.pathname !== wellKnownUrl`.

3. also you don't consider `source` property redundant?

flober81:
Done in 8f72116: not found is just `{"found": false}`, `catalog_url` is only added when the `<link rel="ai-catalog">` target differs from the default path on the page's origin, and `source` is gone since it followed from `catalog_url` anyway. One small deviation: `catalog_url` is kept when that custom target fails, so "announced path is broken" stays distinguishable from "no catalog". The WPT run shows the slim output for all three test sites.

max-ostapenko (approved):

PR #202, opened 2026-09-22
Update PR checkout logic and conditional comment in wpt-test workflow

max-ostapenko:
This pull request updates the `.github/workflows/wpt-test.yml` workflow to improve how pull request changes are checked out and to refine when PR comments are posted. The main goal is to handle internal and forked pull requests more securely and efficiently, and to avoid posting empty comments.

**Workflow improvements:**

* Improved checkout logic:
  - Internal PRs (from the same repository) now perform a full checkout of the branch.
  - Forked PRs only check out the `dist/` directory, reducing the risk of running untrusted code.


tunetheweb (approved):
This is because what we did in https://github.com/HTTPArchive/wappalyzer/pull/331 is overkill for this repo?

max-ostapenko:
@tunetheweb the root is - [the test workflow broke](https://github.com/HTTPArchive/custom-metrics/pull/201#issuecomment-5773063321).
Here is no GCP token, so the surface is smaller.
And as you also noticed, the checks visibility is worse with https://github.com/HTTPArchive/wappalyzer/pull/331.

So kept changes to the minimum here.

